PR #22, opened 2022-03-16
feat: Docker Build + Push; Get Github Ref Names

edulop91:
Add actions to calculate the github base and head ref names from github action payloads. Use this in docker action to calculate caches

### DIFF
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -1,9 +1,6 @@
 name: Create or Update a Happy Stack
 description: "Will create or update a Happy Stack"
 inputs:
-  tfe-token:
-    description: "TFE Token"
-    required: true
   stack-name:
     description: "Name of the stack to create or update"
     required: true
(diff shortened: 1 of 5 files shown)


.github/workflows/get-github-ref-names.yml
```diff
@@ -0,0 +1,24 @@
+on:
```
edulop91, line 1: test for ourselves

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -1,37 +1,38 @@
...
+  tfe-token:
+    description: "TFE Token"
+    required: true
 runs:
   using: "composite"
   steps:
     - uses: actions/checkout@v2
     - name: Install happy
-      uses: chanzuckerberg/github-actions/.github/actions/install-happy@41c7efb56fadc2517f6843a50c1d42bd38a47b9a
+      uses: chanzuckerberg/github-actions/.github/actions/install-happy@9f6d901750728bf97041063580eeda064ff74322
```
alexlokshin-czi, line 19: How do we pull the right version of happy cli?

alexlokshin-czi (approved):

.github/actions/deploy-happy-stack/action.yml
edulop91: there's an argument you can specify, currently defaults to 0.15.3. 
I think I want to make it default to "latest" and instead have users pin a version if they want

PR #90, opened 2022-05-02
fix: conventional-commits can match more characters in () identifier

edulop91:
so we can do things like `feat(FOOBAR-123): do the thing`

### DIFF
--- a/.github/actions/conventional-commits/action.yml
+++ b/.github/actions/conventional-commits/action.yml
@@ -6,7 +6,7 @@ runs:
     - uses: actions/github-script@v5
       with:
         script: |
-          const validator = /^(chore|feat|fix|revert|docs|style)(\([a-z ]+\))?(!)?: (.)+$/
+          const validator = /^(chore|feat|fix|revert|docs|style)(\([a-zA-Z0-9- ]+\))?(!)?: (.)+$/
           const title = context.payload.pull_request.title
           const is_valid = validator.test(title)
 
@@ -17,4 +17,3 @@ runs:
             })
             core.setFailed(`Your pr title doesn't adhere to conventional commits syntax. See more details: ${details}`)
           }
-# bump
--- a/.github/actions/conventional-commits/action.yml
+++ b/.github/actions/conventional-commits/action.yml
@@ -9,7 +9,7 @@ runs:
   steps:
     - uses: actions/setup-node@v3
       with:
-        node-version: "14"
+        node-version: "18"
         check-latest: true
     - run: npm install csv-parse
       shell: bash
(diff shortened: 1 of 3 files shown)


jakeyheath (approved):

.github/actions/conventional-commits/action.yml
```diff
@@ -1,20 +1,78 @@
 name: Conventional Commits Checking
 description: "Validates that a pull request title is compatible with Conventional Commits."
+inputs:
+  enforce_identifier_prefixes:
+    description: "List of JIRA (or similar) project identifiers to enforce. eg: `feat(FOO-123):...` or `feat(BAR-123):...` then pass in FOO\nBAR"
```
jakeyheath, line 5: I will add that this description is kind of confusing. I'm not really sure what it means.

jakeyheath: Can you provide an example of what format you are expecting here. Like is it a string of characters that you are allowing or is it a regexp or something else.

edulop91: https://github.com/chanzuckerberg/github-actions/pull/90/files#diff-e27c2676c8ee63a108d6d91680cbeca9fdce68ff977cc4914b149f7e40b065f0R16

.github/actions/conventional-commits/action.yml
```diff
@@ -1,20 +1,78 @@
...
+      with:
+        node-version: "18"
+        check-latest: true
+    - run: npm install csv-parse
+      shell: bash
     - uses: actions/github-script@v5
       with:
         script: |
-          const validator = /^(chore|feat|fix|revert|docs|style)(\([a-z ]+\))?(!)?: (.)+$/
+          const {parse} = require('csv-parse/sync');
```
jakeyheath, line 19: If we are going to be using a library, is there just a node library that checks this already? all this regex stuff seems very error prone, but you know my feelings about regexp.

jakeyheath: like this? https://github.com/conventional-changelog/commitlint

.github/actions/conventional-commits/action.yml
jakeyheath: ya I guess that example doesn't help me understand the newline character mentioned in this description

.github/workflows/conventional_commits_title.yml
```diff
@@ -13,3 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v3
       - uses: ./.github/actions/conventional-commits
+      - uses: ./.github/actions/conventional-commits
+        with:
+          enforce_identifier_prefixes: |
+            CCIE
```
edulop91, line 19:
```suggestion
            CCIE
            CID
```

.github/workflows/conventional_commits_title.yml
```diff
@@ -13,3 +13,8 @@ jobs:
...
+            CID
```
edulop91, line 20:
```suggestion
            CDI
```

.github/actions/conventional-commits/action.yml
edulop91: you're right

jakeyheath:
stale

PR #92, opened 2022-05-04
fix: ends with for registry validation

jakeyheath:
This PR adds a validation step for the registry input to make sure it doesn't end in a slash. If it does, docker login fails with "unauthorized" which is an error I have encountered a few times now and a total red herring.


jakeyheath:
What would be a good way to test this?

PR #115, opened 2022-06-08
feat: add delete functionality to deploy-happy-stack

lgan-czi:
Background:
The imaging team working on napari-hub wants to deprecate their happy build process and move to using CZI's reusable actions instead. However, the reusable actions currently don't have a way to delete a stack, and we would need to add that functionality to move entirely to using CZI's reusable actions. This change adds additional flexibility for teams managing their stack resources.
Additionally, while testing our changes, we ran into an error where the script wouldn't create a new stack if there was another stack with a name that contained the new stack's name. For example, if we wanted to create a stack called test but we had already made a stack called test-npm, the script would only try to update test because grep had found an occurrence of test in test-npm in the output of happy list.

Changes:
We added a delete functionality to deploy-happy-stack in order to allow us to migrate over fully. To delete a stack, users should set the delete-stack input to true, which would run the happy delete command under the if statement: `if ${DELETE_STACK}; then`, deleting the stack.
We also modified the regex to be more accurate in getting the stack name in the if statement. We're now looking for the pattern `"^|\s*${STACK_NAME}\s*|"` instead of just `${STACK_NAME}`so we can avoid the issue we ran into above. The pattern `"^|\s*${STACK_NAME}\s*|"` looks for a pattern at the beginning of each line that has the STACK_NAME sandwiched between whitespaces and two pipes, which should be able to more accurately check for existence of the correct stack name instead of another stack name that happens to contain that stack name. One remaining issue is that the script won't work if we try to create a happy stack was named NAME because `happy list` will always output NAME as part of the table it creates.

Testing Approach:
The additional if statement was initially tested locally by running a bash script on a mac and ensuring the correct block ran based on whether the DELETE_STACK was true or false. 
We then tested deployments in our branch using this fork until the script was able to create, update, and delete stacks.

### DIFF
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -27,7 +27,7 @@ inputs:
     description: "The environment to update or create"
     required: false
   flags:
-    description: "Additional flags to set for happy commands. Flags --aws-profile "" and --env=${ENV} are currently always used."
+    description: "Additional flags to set for happy commands. Flags --aws-profile  and --env=${ENV} are currently always used."
     default: ""
     required: false
 
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -54,7 +54,7 @@ runs:
         if ${DELETE_STACK}; then
           echo "Deleting stack ${STACK_NAME}"
           happy delete --aws-profile "" --env=${ENV} ${STACK_NAME}
-        elif $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q "^${STACK_NAME}$" ); then
+        elif $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q ${STACK_NAME} ); then
           echo "Updating stack ${STACK_NAME}"
           happy update --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
         else
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -56,7 +56,7 @@ runs:
         if ${DELETE_STACK}; then
           echo "Deleting stack ${STACK_NAME}"
           happy delete --aws-profile "" --env=${ENV} ${STACK_NAME}
-        elif $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q "^|\s*${STACK_NAME}\s*|" ); then
+        elif $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q "^\s*\|\s*${STACK_NAME}\s*\|" ); then
           echo "Updating stack ${STACK_NAME}"
           happy update --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
         else
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -1,6 +1,10 @@
 name: Create or Update a Happy Stack
 description: "Will create, update, or delete a Happy Stack"
 inputs:
+  operation:
+    description: "Operation we want this to perform. Defaults to create-or-update. Set to delete if you want to delete the stack."
+    default: ""
+    required: false
   stack-name:
     description: "Name of the stack to create, update, or delete"
     required: true
@@ -22,10 +26,6 @@ inputs:
   env:
     description: "The environment to update or create"
     required: false
-  delete-stack:
-    description: "Should we delete current stack. Defaults to false."
-    default: false
-    required: false
 
 runs:
   using: "composite"
@@ -42,7 +42,7 @@ runs:
         CREATE_TAG: ${{ inputs.create-tag }}
         TAG: ${{ inputs.tag }}
         ENV: ${{ inputs.env }}
-        DELETE_STACK: ${{ inputs.delete-stack }}
+        OPERATION: ${{ inputs.operation }}
       shell: bash
       run: |
         set -ue
@@ -50,14 +50,16 @@ runs:
         
         IMAGE_TAG="${TAG:-sha-$GITHUB_SHA}"
         echo "Targetting Stack ${STACK_NAME}, tagging with '${IMAGE_TAG}'"
+        HAPPY_OPERATION=${OPERATION}
+        HAPPY_DEPLOYMENT_FLAGS=""
 
-        if ${DELETE_STACK}; then
-          echo "Deleting stack ${STACK_NAME}"
-          happy delete --aws-profile "" --env=${ENV} ${STACK_NAME}
-        elif $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q "^|\s*${STACK_NAME}\s*|" ); then
-          echo "Updating stack ${STACK_NAME}"
-          happy update --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
-        else
-          echo "Creating stack ${STACK_NAME}"
-          happy create --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
-        fi
+        if [[ ${HAPPY_OPERATION} != "create" && ${HAPPY_OPERATION} != "update" && ${HAPPY_OPERATION} != "delete" ]]; then
+          HAPPY_DEPLOYMENT_FLAGS="--create-tag=${CREATE_TAG} --tag ${IMAGE_TAG}"
+          if $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q "^|\s*${STACK_NAME}\s*|" ); then
+            HAPPY_OPERATION="update"
+          else
+            HAPPY_OPERATION="create"
+          fi
+        echo "Starting to ${OPERATION} stack ${STACK_NAME} with additional flags: ${HAPPY_DEPLOYMENT_FLAGS}"
+        echo "happy ${OPERATION} --aws-profile "" --env=${ENV} ${HAPPY_DEPLOYMENT_FLAGS} ${STACK_NAME}"
+        happy ${OPERATION} --aws-profile "" --env=${ENV} ${HAPPY_DEPLOYMENT_FLAGS} ${STACK_NAME}
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -51,8 +51,6 @@ runs:
         IMAGE_TAG="${TAG:-sha-$GITHUB_SHA}"
         echo "Targetting Stack ${STACK_NAME}, tagging with '${IMAGE_TAG}'"
 
-        happy list --aws-profile "" --env=${ENV}
-
         if ${DELETE_STACK}; then
           echo "Deleting stack ${STACK_NAME}"
           happy delete --aws-profile "" --env=${ENV} ${STACK_NAME}
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -26,10 +26,6 @@ inputs:
   env:
     description: "The environment to update or create"
     required: false
-  flags:
-    description: "Additional flags to set for happy commands. Flags --aws-profile  and --env=${ENV} are currently always used."
-    default: ""
-    required: false
 
 runs:
   using: "composite"
@@ -47,7 +43,6 @@ runs:
         TAG: ${{ inputs.tag }}
         ENV: ${{ inputs.env }}
         OPERATION: ${{ inputs.operation }}
-        FLAGS: ${{ inputs.flags }}
       shell: bash
       run: |
         set -ue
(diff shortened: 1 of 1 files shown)


.github/actions/deploy-happy-stack/action.yml
```diff
@@ -46,7 +51,12 @@ runs:
         IMAGE_TAG="${TAG:-sha-$GITHUB_SHA}"
         echo "Targetting Stack ${STACK_NAME}, tagging with '${IMAGE_TAG}'"
 
-        if $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q ${STACK_NAME} ); then
+        happy list --aws-profile "" --env=${ENV}
```
potating-potato, line 54: remove this testing line

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -46,7 +51,12 @@ runs:
...
+
+        if ${DELETE_STACK}; then
+          echo "Deleting stack ${STACK_NAME}"
+          happy delete --aws-profile "" --env=${ENV} ${STACK_NAME}
+        elif $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q " ${STACK_NAME} " ); then
```
potating-potato, line 59: add the grep expression to the description as well, including the reason why it was changed

.github/actions/deploy-happy-stack/action.yml
lgan-czi: done!

.github/actions/deploy-happy-stack/action.yml
lgan-czi: done!

potating-potato (approved):
LGTM

alexlokshin-czi (approved):

potating-potato:
@alexlokshin-czi thanks for the quick review! May I ask about the merge and release process here? Right now we are using a commit hash in our workflow, would love to see this released soon

jakeyheath:
Thanks for the contribution! And thanks so much for the detail description. 

Since this action is called deploy-happy-stack, I wonder if we should change the name of the action or roll these feature in a separate github action. The delete feature doesn't really suggest as part of a deployment process to me. I am thinking an action called "delete-stach-if-exists" or something might be more accurate.

jakeyheath:
As for the release process, after this PR is merged, we'll create a release and you should be able to use the new version as soon as the release is created 👍🏻 . We don't have a formal cycle that we follow per say.

jakeyheath (changes requested):
Thanks for the contribution! And thanks so much for the detail description.

Since this action is called deploy-happy-stack, I wonder if we should change the name of the action or roll these feature in a separate github action. The delete feature doesn't really suggest as part of a deployment process to me. I am thinking an action called "delete-stach-if-exists" or something might be more accurate.

As for the release process, after this PR is merged, we'll create a release and you should be able to use the new version as soon as the release is created 👍🏻 . We don't have a formal cycle that we follow per say.

Also, for my own curiosity, why do you need to delete stacks as part of your Github Action CICD? I hadn't thought of that use case before and would love to know how you are using this delete feature.

alexlokshin-czi:
> Thanks for the contribution! And thanks so much for the detailed description.
> 
> Since this action is called deploy-happy-stack, I wonder if we should change the name of the action or roll these features in a separate github action. The delete feature doesn't really seem to fit to me as part of a deployment process. I am thinking an action called "delete-stack-if-exists" or something might be more accurate.
> 
> As for the release process, after this PR is merged, we'll create a release and you should be able to use the new version as soon as the release is created 👍🏻 . We don't have a formal cycle that we follow per say.
> 
> Also, for my own curiosity, why do you need to delete stacks as part of your Github Action CICD? I hadn't thought of that use case before and would love to know how you are using this delete feature.

One thing I'd done differently is re-use the `STACK_NAME`, but introduce an `OPERATION`. We already use `create` and `update` interchangeably - `delete` is a one off use case, that can happen exactly once. Jake's concern, I'd imagine, is why do we need to `delete` through a github action?

potating-potato:
Happy to explain it:
For the imaging team, we have 2 kinds of happy usage. Our staging/prod deployment workflow is probably more aligned to what you already have and use. That is: Update when necessary (maybe a one time creation at the beginning, even though we created the stack manually at the beginning).

The second use case is that we use happy to deploy remote development environment, we trigger the creation, update, and deletion based on branch activity, see [napari hub rdev workflow](https://github.com/chanzuckerberg/napari-hub/blob/main/.github/workflows/push-rdev.yml)

in this workflow, we would delete the remote deployment environment when the branch is deleted (usually as a result of the branch merged to the main branch via squash merge and branch deletion). So to our engineers, the only thing they would need to worry about is git branch. As a result, happy, and essentially terraform all hide behind the scenes and maintain the stack without anyone knowing the technical specifics. By doing this we simplified the flow for our engineers and saved most people on the team from having to know happy existence/how-to-use, and their development cycle is just push-to-branch -> get-URL-to-inspect-the-change -> merge-or-delete-branch

jakeyheath:
> in this workflow, we would delete the remote deployment environment when the branch is deleted (usually as a result of the branch merged to the main branch via squash merge and branch deletion). 
ooo this is a super great workflow actually. We've had a few instances where folks have a bunch of stale stacks lingering around and this would help a lot with that workflow, I think.

I agree with @alexlokshin-czi . Can we change the delete-stack flag to an OPERATION input that is either "create-or-update" by default or "delete". If delete, do the delete flow. If create-or-update, do the create/update flow.

potating-potato:
ok, we are reworking the PR and will unify the usage pattern under the operations flag while maintaining backward compatibility

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -38,18 +42,25 @@ runs:
...
+        HAPPY_OPERATION=${OPERATION}
+        HAPPY_DEPLOYMENT_FLAGS=""
 
-        if $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q ${STACK_NAME} ); then
-          echo "Updating stack ${STACK_NAME}"
-          happy update --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
-        else
-          echo "Creating stack ${STACK_NAME}"
-          happy create --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
+        if [[ ${HAPPY_OPERATION} != "create" && ${HAPPY_OPERATION} != "update" && ${HAPPY_OPERATION} != "delete" ]]; then
```
potating-potato, line 56: should just check for emptiness here, if it is set then we should respect the operation given

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -38,18 +46,26 @@ runs:
...
+        OPERATION: ${{ inputs.operation }}
+        FLAGS: ${{ inputs.flags }}
       shell: bash
       run: |
         set -ue
         set -o pipefail
         
         IMAGE_TAG="${TAG:-sha-$GITHUB_SHA}"
         echo "Targetting Stack ${STACK_NAME}, tagging with '${IMAGE_TAG}'"
+        HAPPY_OPERATION=${OPERATION}
```
potating-potato, line 58: would be simpler just use OPERATION later? any benefit in reassigning another variable to it?

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -22,6 +26,10 @@ inputs:
   env:
     description: "The environment to update or create"
     required: false
+  flags:
```
potating-potato, line 29: For now I would rather not add support for additional flags, unless other people think otherwise we should drop these

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -38,18 +46,26 @@ runs:
...
+        if [[ ${HAPPY_OPERATION} == "" ]]; then
+          HAPPY_DEPLOYMENT_FLAGS="--create-tag=${CREATE_TAG} --tag ${IMAGE_TAG}"
+          if $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q "^|\s*${STACK_NAME}\s*|" ); then
+            HAPPY_OPERATION="update"
+          else
+            HAPPY_OPERATION="create"
+          fi
         fi
+        echo "Starting to ${HAPPY_OPERATION} stack ${STACK_NAME} with additional flags: ${HAPPY_DEPLOYMENT_FLAGS}"
+        echo "happy ${HAPPY_OPERATION} --aws-profile "" --env=${ENV} ${HAPPY_DEPLOYMENT_FLAGS} ${STACK_NAME}"
```
potating-potato, line 70: I don't think we would want to print this in a production environment, the above message should be sufficient.

lgan-czi: Yep, will remove once I'm done testing

.github/actions/deploy-happy-stack/action.yml
lgan-czi: Sounds good!

.github/actions/deploy-happy-stack/action.yml
lgan-czi: I thought we couldn't change those variable for some reason, it's been fixed

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -1,8 +1,12 @@
 name: Create or Update a Happy Stack
-description: "Will create or update a Happy Stack"
+description: "Will create, update, or delete a Happy Stack"
 inputs:
+  operation:
+    description: "Operation we want this to perform. Defaults to create-or-update. Set to delete if you want to delete the stack."
```
potating-potato, line 5: maybe change the default description to "If not provided, happy will automatically create or update the stack with given name"

potating-potato:
@alexlokshin-czi @jakeyheath We are still in the middle of polishing the change, can you please give a cursory look at whether you like the direction here? Thanks!

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -1,8 +1,12 @@
...
+    default: ""
```
jakeyheath, line 6:
```suggestion
    default: "create-or-update"
```

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -38,18 +42,24 @@ runs:
...
         echo "Targetting Stack ${STACK_NAME}, tagging with '${IMAGE_TAG}'"
+        HAPPY_DEPLOYMENT_FLAGS=""
 
-        if $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q ${STACK_NAME} ); then
-          echo "Updating stack ${STACK_NAME}"
-          happy update --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
-        else
-          echo "Creating stack ${STACK_NAME}"
-          happy create --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
+        if [[ ${OPERATION} == "" ]]; then
```
jakeyheath, line 55:
```suggestion
        if [[ ${OPERATION} == "create-or-update" ]]; then
```

.github/actions/deploy-happy-stack/action.yml
lgan-czi: Just updated!

.github/actions/deploy-happy-stack/action.yml
jakeyheath: I think it reads better to use the actual string that describes the operation rather than an empty string.

jakeyheath (approved):
LGTM

lgan-czi:
Thanks for all the reviews and comments! I just finished doing a few last minute checks, and it looks like everything is working. Can someone merge these changes? @alexlokshin-czi @jakeyheath

.github/actions/deploy-happy-stack/action.yml
```diff
@@ -38,18 +42,26 @@ runs:
...
+        HAPPY_DEPLOYMENT_FLAGS=""
 
-        if $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q ${STACK_NAME} ); then
-          echo "Updating stack ${STACK_NAME}"
-          happy update --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
-        else
-          echo "Creating stack ${STACK_NAME}"
-          happy create --aws-profile "" --env=${ENV} --create-tag=${CREATE_TAG} --tag ${IMAGE_TAG} ${STACK_NAME}
+        if [[ ${OPERATION} == "create-or-update" ]]; then
+          HAPPY_DEPLOYMENT_FLAGS="--create-tag=${CREATE_TAG} --tag ${IMAGE_TAG}"
```
potating-potato, line 56: nit picking: if this is set below anyway, why not change the condition check below to if from elif, so that only one place sets the flags?

lgan-czi: My intent was to make the distinction between what to do if an operation input was given and if an input wasn't given. Only setting the flags in one place feels better though. Will change.

jakeyheath:
I've merged this in, we can always iterate on it to make it better. Hopefully in the future, we can get away from all this bash because I think in general is a little confusing to add more and more complex features to it, but I think this is a great contribution. Thanks so much. I will make a release and you should be able to reference the tag.

jakeyheath:
Cool, it is now released!

Please use the [tag](https://github.com/chanzuckerberg/github-actions/releases/tag/deploy-happy-stack-v1.4.0) `deploy-happy-stack-v1.4.0` in your ref to grab that change in your CICD.

PR #129, opened 2022-06-30
feat: (CCIE-308) Adding delete action

alldoami:
We are wanting to delete branches, so this will cover that action.

Tested it in our new playground repo: https://github.com/chanzuckerberg/github-actions-playground/runs/7139716330?check_suite_focus=true

I believe these are the expected results. @jakeyheath lmk if this is right! We're wanting to get the branch name of branch that was just deleted right?

### DIFF
--- a/.github/actions/get-github-ref-names/action.yml
+++ b/.github/actions/get-github-ref-names/action.yml
@@ -30,7 +30,7 @@ runs:
           if (eventName === "pull_request") {
             head = context.payload.pull_request.head.ref;
             base = context.payload.pull_request.base.ref;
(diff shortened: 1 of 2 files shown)


jakeyheath:
Cool! Were you able to test this by chance? Or can you provide a reference to the github action docs that shows this would work?

jakeyheath (approved):

alldoami:
@jakeyheath I'm going to test it tomorrow! This was just by first thoughts...

jakeyheath:
Looks great!

edulop91 (approved):

PR #130, opened 2022-06-30
fix: Stack update fails as grep doesn't pick up the output

alexlokshin-czi:


.github/actions/deploy-happy-stack/action.yml
```diff
@@ -57,7 +57,7 @@ runs:
         HAPPY_DEPLOYMENT_FLAGS=""
 
         if [[ ${OPERATION} == "create-or-update" ]]; then
-          if $(happy list --aws-profile "" --env=${ENV} 2>&1 | grep -q "^|\s*${STACK_NAME}\s*|" ); then
+          if $(happy list --aws-profile "" --env=${ENV} &>out.txt; cat out.txt | grep "^|\s*${STACK_NAME}\s*|" ); then
```
jgadling, line 60: I think `grep 'search_string' out.txt` is canonical here.

jgadling: also -- wow. most `happy` commands should probably have a json output option so we can find the data we're looking for with `jq` -- this is error prone (we could match the table header row for example)

jgadling (approved):
I'm not sure why this broke, or why this PR fixes it. did the `happy` command fail in the case of the action that tried to recreate a stack? Does this prevent that failure mode?

PR #137, opened 2022-07-13
fix: When happy list errors out for whatever reason, treat it as a failure, not as an update

alexlokshin-czi:


.github/actions/deploy-happy-stack/action.yml
```diff
@@ -57,7 +57,16 @@ runs:
...
+          happy list --aws-profile "" --env=${ENV} &>out.txt
+          EXIT_CODE=$?
+          if [[ $EXIT_CODE -ne 0 ]]; then
+              cat out.txt
+              exit $EXIT_CODE
+          fi
+
+          STACKS=$(grep "^|\s*${STACK_NAME}\s*|" out.txt)
+         
+          if [[ $STACKS ]]; then
```
jgadling, line 69: should this be an `if [[ -n $STACKS ]]; then` ? My bash is terrible though

alexlokshin-czi: I've changed it to `[[ -n "${STACKS}" ]]` as you requested, but `[[ ! -z "${STACKS}" ]]` and `[[ $STACKS ]];` behave identically with the rest of the script -- verified locally with bash and sh.

jgadling (approved):

PR #160, opened 2023-01-25
fix: find correct latest happy version

hspitzley-czi:
First part of https://czi-tech.atlassian.net/browse/CCIE-1022


jakeyheath (approved):
Looks great! After this, can you update the actions in this repo that depend on this action?

PR #170, opened 2023-05-04
feat: enable optional docker-compose-path argument

shitchcock:
Description:

I'm trying to use a custom docker compose config path with my happy setup. Would like to bring this possibility to the shared GHA

### DIFF
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -29,6 +29,9 @@ inputs:
   env-file:
     description: "Specify a docker env file to use when deploying"
     required: false
+  docker-compose-config-path:
+    description: "Specify a docker compose config path"
+    required: false
   skip-migrations:
     description: "Specify true to skip migrations"
     default: "false"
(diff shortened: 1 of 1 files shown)


.github/actions/deploy-happy-stack/action.yml
```diff
@@ -54,6 +54,7 @@ runs:
         TAG: ${{ inputs.tag }}
         ENV: ${{ inputs.env }}
         ENV_FILE: ${{ inputs.env-file }}
+        DOCKER_COMPOSE_PATH: ${{ inputs.docker-compose-path }}
```
jakeyheath, line 57: I think you will also need to add an input at the top of this file: 

https://github.com/chanzuckerberg/github-actions/blob/0974ca9a40d1fc500a675e4d7fff8bd307655ce5/.github/actions/deploy-happy-stack/action.yml#L3

also, to make it a little more clear, can we call it `docker-compose-config-path`

jakeyheath (changes requested):

.github/actions/deploy-happy-stack/action.yml
shitchcock: Good catch thank you!

jakeyheath (approved):

PR #184, opened 2023-06-16
feat: new composite action to validate jira issue references in PRs

hspitzley-czi:


.github/actions/jira-find-marker/action.yaml
```diff
@@ -0,0 +1,48 @@
+name: Matched Jira Marker
```
hspitzley-czi, line 1: Moved from eng-portal so we can re-use in happy

.github/actions/jira-validate-reference/action.yaml
```diff
@@ -0,0 +1,175 @@
+name: Validate Jira Reference
```
hspitzley-czi, line 1: Moved from eng-portal so we can re-use in happy

.github/actions/jira-validate-reference/action.yaml
```diff
@@ -0,0 +1,175 @@
...
+            if (match) {
+              jiraIssueRef = match[0];
+              console.log(`Found Jira issue reference in PR ${label}: ${jiraIssueRef}`);
+              return true;
+            }
+            return false;
+          });
+
+          if (!jiraIssueRef) {
+            throw new Error(`No Jira issue reference found in branch, title, or body of PR. Please add a reference to a Jira issue in the form of ${{ inputs.projectKey }}-#### to the branch name, title, or body of your PR.`);
```
alexlokshin-czi, line 85: Maybe, an example of `CCIE-1400` would be appropriate here.

PR #206, opened 2023-08-07
feat: CCIE-1217: Read the .happy/version.lock file for happy version

dspadea:
Get Happy version from the project's `.happy/version.lock` file if available and no specific version is specified in inputs.

### DIFF
--- a/.github/actions/install-happy/action.yaml
+++ b/.github/actions/install-happy/action.yaml
@@ -42,7 +42,7 @@ runs:
         GHCLI_TOOL_PATH="/tmp/ghcli"
         mkdir -p $GHCLI_TOOL_PATH
         cd /tmp/
-        curl $URL -L -v -o ghcli.tar.gz 
+        curl $URL -L -v -o ghcli.tar.gz
         tar -xf ghcli.tar.gz -C $GHCLI_TOOL_PATH --strip-components 1
         echo "${GHCLI_TOOL_PATH}/bin" >> "${GITHUB_PATH}"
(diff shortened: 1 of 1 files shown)


.github/actions/install-happy/action.yaml
```diff
@@ -87,8 +87,13 @@ runs:
         set -o pipefail
         cd $(mktemp -d)
         echo Input argument indicated to use Happy version: $HAPPY_VERSION
-        if [ -z $HAPPY_VERSION ] || [ $HAPPY_VERSION == "latest" ]
+        if [ -z $HAPPY_VERSION ] && [ -e ".happy/version.lock" ]
         then
+          echo "Happy version was not specified, but project has a .happy/version.lock. Getting version from there."
```
hspitzley-czi, line 92: Two things:
1. I just pushed a fix to the latest version selection that re-writes a lot of this step in JS so I think I caused a lot of merge conflicts for you here
2. I think passing an arg `happy-version-file` would be more flexible and allow using this feature in monorepos where the happy root is not the same as the repo root (this would be similar to how [setup-go](https://github.com/actions/setup-go) has a go-version and a go-version-file arg, [setup-node](https://github.com/actions/setup-node) is designed that way too). Thoughts?

dspadea: Ok, I can add that. I'll figure out how to work it into the JS. Haven't looked at it yet.

dspadea:
Re-wrote in javascript per above. Closing this PR to #211 .

PR #207, opened 2023-08-09
fix: ensure happy asset exists when selecting latest release

hspitzley-czi:
When choosing the latest happy release we need to check that the asset we want to install has been added by goreleaser. Without this we have seen intermittent failures. re: https://czi-tech.atlassian.net/browse/CCIE-1653

This code uses the following process to choose the happy version to use:
1. Get a list of releases (ordered newest to oldest)
2. Iterate through the releases and when is encountered that has the tag format `v#.#.#` then check for the asset that we need to install
   - If the asset is found: use that version
   - If the asset is not found: continue to the next-newest version with the correct tag format

### DIFF
--- a/.github/actions/install-happy/action.yaml
+++ b/.github/actions/install-happy/action.yaml
@@ -81,6 +81,7 @@ runs:
       with:
         releaseNameRegEx: '^v[0-9]+\.[0-9]+\.[0-9]+$'
     - name: Print releases
+      shell: bash
       run: |
         echo "Releases: ${{ steps.getReleases.outputs }}"
 
--- a/.github/actions/install-happy/action.yaml
+++ b/.github/actions/install-happy/action.yaml
@@ -76,7 +76,7 @@ runs:
         rm -rf /tmp/aws/
 
     - name: Determine happy version to use
-      id: getReleases
+      id: determineLatestHappyCli
       uses: actions/github-script@v6
       with:
         script: |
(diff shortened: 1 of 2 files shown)


.github/actions/install-happy/action.yaml
```diff
@@ -86,23 +142,8 @@ runs:
...
-        ARM_CMD="wget --quiet https://github.com/chanzuckerberg/happy/releases/download/v${HAPPY_VERSION}/happy_${HAPPY_VERSION}_linux_arm64.tar.gz -O happy.tar.gz"
-        AMD_CMD="wget --quiet https://github.com/chanzuckerberg/happy/releases/download/v${HAPPY_VERSION}/happy_${HAPPY_VERSION}_linux_amd64.tar.gz -O happy.tar.gz"
-        if [[ $PLATFORM == "arm64" ]]; then
-            $($ARM_CMD)
-        elif [[ $PLATFORM == "aarch64" ]]; then
-            $($ARM_CMD)
-        else
-            $($AMD_CMD)
-        fi
+        echo Installing Happy from asset: $HAPPY_RELEASE_ASSET
```
alexlokshin-czi, line 145:
```suggestion
        echo Installing Happy from version $HAPPY_VERSION, asset: $HAPPY_RELEASE_ASSET
```

alexlokshin-czi (approved):

PR #209, opened 2023-08-09
feat: Allow the github ref specification

alexlokshin-czi:


.github/actions/deploy-happy-stack/action.yml
```diff
@@ -49,11 +48,16 @@ inputs:
...
+  github-repo-branch:
+    description: "Branch of Github repo to deploy"
+    required: false
 
 runs:
   using: "composite"
   steps:
     - uses: actions/checkout@v3
+      with:
+          ref: ${{ inputs.github-repo-branch || github.event.pull_request && github.head_ref || github.ref_name }}
```
jfoo1984, line 60: For context, the reason this isn't more straightfoward is here: https://github.com/github/docs/issues/15319#issuecomment-1620624403

jfoo1984:
I'm not sure the best place to raise this, but it may also be helpful to document that this action performs its own checkout.  That wasn't clear to me until I looked into the code for the action

jakeyheath (approved):

PR #211, opened 2023-08-14
feat: CIE-1217: If no happy version specified, check version.lock

dspadea:


.github/actions/install-happy/action.yaml
```diff
@@ -95,7 +96,36 @@ runs:
...
             return;
           }
-          
+
+          if (! process.env.HAPPY_VERSION ) {
+
+            var version_lock_file = ".happy/version.lock";
+            if (process.env.VERSION_LOCK_FILE ) {
+              version_lock_file = process.env.VERSION_LOCK_FILE
+            }
```
hspitzley-czi, line 105: I don't think this will ever evaluate to true since we default the happy version to "latest". Maybe this should just check if the happy version file arg is specified (without trying to default it to the default location)?
```suggestion
          if (process.env.VERSION_LOCK_FILE) {
```

hspitzley-czi: I'm good with that 👍 

You'll need to remove the default here as well (and make it not required) https://github.com/chanzuckerberg/github-actions/blob/aa19edd466d4869763a06f944eb922596b74f6d7/.github/actions/deploy-happy-stack/action.yml#L24-L25

PR #212, opened 2023-08-23
fix: mismatched syntax for jq command in happy cleanup

shitchcock:
The command 
```
happy list --aws-profile "" --output json --env $ENV | jq -r --arg date "$date" '.[] | select(.LastUpdated < $date) | .Name'
```
Should return a list of names of happy stacks from the json output. Instead it returns Nulls.

Updating to 

```
happy list --aws-profile "" --output json --env $ENV | jq -r --arg date 2023-08-23T22:08:49+00:00 '.[] | select(.last_updated < $date) | .stack'
```
returns the expected outcome

### DIFF
--- a/.github/actions/happy-cleanup/action.yml
+++ b/.github/actions/happy-cleanup/action.yml
@@ -35,7 +35,7 @@ runs:
         set -o pipefail
 
         date=`date +%Y-%m-%d'T'%H:%M'Z' -d "$TIME ago"`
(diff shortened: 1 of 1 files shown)


forgeabout (approved):
Yeah, looks like Happy changed its output format (or the cleanup script was always wrong)? Your new field names look good to me.

jakeyheath (approved):

jakeyheath (approved):

jakeyheath:
I think this was a change in that we now query the API for the stack list and not running the query locally. LGTM!

PR #213, opened 2023-08-24
feat: terraform plan github action

jakeyheath:
## Summary

This PR introduces a new action in the works. I think it needs a few more iterations, but as is we could start experimenting. The workflow would allow us to execute terraform plans and validations bases on PRs. Right now, we have a very tight coupling mechanism between TFE and PRs. It takes a very long time to run all the speculative plans and prevents us from making out of band terraform applies and plans. Generally having terraform workspaces tied to a git SHA is helpful, but for the core infra team it is a clunky workflow. 

This is the first step to changing this workflow. Here is what it would look like:

* All TFE workspaces would stay the same
* We would update the VCS settings to be CLI driven
* All PRs would run this tf-plan action based on the files that have changed
* Plans would be printed for all workspaces in the PR with their corresponding links in TFE
* Plans would be applied when pushed to main

This is essentially the same workflow we have now with a few small changes:

* Handle dependencies between workspaces within the same PR (right now, you have to make several PRs if you have dependencies)
* It allows us to leverage our autoscaling Github Action runners (TFE doesn't autoscale right now and is still pretty slow)
* It allows us to speed up our PR approval steps for terraform
* First step toward remove TFE, but for now all plans will be executed against TFE (in the future, we could point them at an S3 bucket and it would work the same)
* Make TF plans/applies out of band
* Cacheing will speed up workflows a lot (TFE doesn't seem to cache right now or at least it still feels very slow)

Here is an example of the PR messaging:

Screenshot 2023-08-28 at 9 07 48 AM

## Items Left to Do

* Set up caching properly
* Attempt to execute with terragrunt instead of terraform (would handle dependencies between workspaces NO MORE MULTI-PR workflow!)
* Build into fogg

### DIFF
--- a/.github/workflows/test-tf-plan.yml
+++ b/.github/workflows/test-tf-plan.yml
@@ -1,12 +1,10 @@
-on: 
-  push: {}
+on:
   pull_request:
-      types:
-      - opened
+    types:
       - edited
+      - opened
+      - synchronize
       - reopened
-      branches:
-      - fogg-*
 
 jobs:
   find-changed-dirs:
--- a/.github/workflows/test-tf-plan.yml
+++ b/.github/workflows/test-tf-plan.yml
@@ -53,5 +53,5 @@ jobs:
       with:
         ssh-private-key: ${{secrets.SHARED_INFRA_DEPLOY_KEY}}
         working-dir: ${{matrix.workspace}}
-        tf-version: ${{steps.get-tf-version-fogg.outputs.version}}
+        tf-version: 1.5.2
         tfe-token: ${{secrets.TFE_TOKEN}}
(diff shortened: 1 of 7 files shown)


.github/workflows/test-tf-plan.yml
```diff
@@ -0,0 +1,55 @@
+on:
+  pull_request:
+    types:
+      - edited
+      - opened
+      - synchronize
+      - reopened
+
+jobs:
+  find-changed-dirs:
```
jakeyheath, line 10: for projects that were really large (like shared-infra), this might not work. We might have to have static jobs or something (I think there is a limit on the number of matrix items you can have)

alexlokshin-czi: We need to break shared-infra up.

alexlokshin-czi (approved):

PR #219, opened 2023-08-29
feat: exclude stacknames

shitchcock:
I would like to add a feature that allows stacknames to be excluded from deletion. 

i.e. edu labels our stacks with `-stable` if they are the stable deployed version of that stack. I would like to exempt these stacks from autodelete

### DIFF
--- a/.github/actions/happy-cleanup/action.yml
+++ b/.github/actions/happy-cleanup/action.yml
@@ -9,6 +9,9 @@ inputs:
     description: The longest period before a stack should be considered stale. Should be something like '2 weeks' or '1 day'
     required: true
     default: 2 weeks
+  exclude:
+    description: Exclude stacks with these substrings in the stackname
+    required: false
   tfe_token:
     description: A Terraform Enterprise API token to be used with your happy organization
     required: true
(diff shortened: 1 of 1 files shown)


.github/actions/happy-cleanup/action.yml
```diff
@@ -29,12 +32,19 @@ runs:
...
         set -o pipefail
 
         date=`date +%Y-%m-%d'T'%H:%M'Z' -d "$TIME ago"`
+        if [[ ! -z ${EXCLUDE} ]]; then
+          for i in $(happy list --aws-profile "" --output json --env $ENV | jq -r --arg date "$date" --arg exclude "$EXCLUDE" '.[] | select(.last_updated < $date) | select(any(.stack; contains($exclude))|not) | .stack'); do
+            happy delete $i --env $ENV --aws-profile ""
+          done
+          exit
+        fi
         for i in $(happy list --aws-profile "" --output json --env $ENV | jq -r --arg date "$date" '.[] | select(.last_updated < $date) | .stack'); do
```
jakeyheath, line 48: do you need to delete this for loop? otherwise I think it will still do the same behavior as before

jakeyheath: also I'm not a super good reviewer of bash scripts, it looks like it would work. Might be a good idea to make a test branch in the happy repo and setup a test with the examples folder.

jakeyheath: or maybe it would just be faster to merge and try it out 🤷🏻 . up to you!

shitchcock: @jakeyheath I don't need to delete the second forloop because it should exit before the second loop executes.

I can set up a test branch to test this based of the sha

shitchcock: In looking into setting up test branches for this I am wondering if it would be faster to merge and test. I cleaned up many of the superfluous stacks last week so I don't have a big pool to pull from but I can demonstrate the outputs of these commands in the test plan

shitchcock: I was able to test the command and am confident it will work

jakeyheath (approved):

PR #220, opened 2023-08-30
fix: CCIE-1770: Add inputs.version_lock_file

dspadea:
Add missing input parameter `version_lock_file`

### DIFF
--- a/.github/actions/install-happy/action.yaml
+++ b/.github/actions/install-happy/action.yaml
@@ -17,6 +17,10 @@ inputs:
     description: Whether to add happy to the system bin directory
     required: false
     default: ""
+  version_lock_file:
+    description: Path to version.json file
+    required: false
+    default: ".happy/version.lock"
 runs:
   using: "composite"
   steps:


alexlokshin-czi (approved):

.github/actions/install-happy/action.yaml
```diff
@@ -17,6 +17,10 @@ inputs:
     description: Whether to add happy to the system bin directory
     required: false
     default: ""
+  version_lock_file:
```
hspitzley-czi, line 20: Could you also add this to https://github.com/chanzuckerberg/github-actions/blob/main/.github/actions/deploy-happy-stack/action.yml and pass it through since that's how the install-happy action is often invoked?

dspadea: Good idea! I'll enter a ticket for it.

dspadea:
https://czi-tech.atlassian.net/browse/CCIE-1771 added to track adding this to `deploy-happy-stack` action

PR #233, opened 2023-12-19
feat: Add resusable action to set app config with happy config

jfoo1984:
Add a new github action to set app configs with happy config


jfoo1984:
Any workflow / step that uses this action needs to have permission to write the id-token
```
permissions:
  id-token: write
```

Is there a way to document or enforce this in the action?

jfoo1984:
I originally developed this in czid-graphql-federation-server/pull/52.  If we end up merging it here, I'll switch that repo to use this action instead.

jakeyheath (approved):

PR #236, opened 2024-01-04
feat: add slice to happy deploy stack action

aldengolab:
Adds a slice option to the happy stack deploy action.

### DIFF
--- a/.github/actions/deploy-happy-stack/action.yml
+++ b/.github/actions/deploy-happy-stack/action.yml
@@ -64,6 +64,9 @@ inputs:
   github-repo-branch:
     description: "Branch of Github repo to deploy"
     required: false
+  slice:
+    description: "Slice to deploy"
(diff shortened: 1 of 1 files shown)


.github/actions/deploy-happy-stack/action.yml
```diff
@@ -64,6 +64,9 @@ inputs:
   github-repo-branch:
     description: "Branch of Github repo to deploy"
     required: false
+  slice:
```
jakeyheath, line 67: Can we also add space for `--slice-default-tag`? We might need this if I recall to change the default tags used by the other services when using a slice option. I'm a little fuzzy on the details, but wouldn't hurt to add

aldengolab: added!

jakeyheath (approved):

PR #262, opened 2024-04-30
feat: add working-dir to happy cleanup

shitchcock:
Adding working directory option for the monorepo


.github/actions/happy-cleanup/action.yml
```diff
@@ -15,6 +15,10 @@ inputs:
   tfe_token:
     description: A Terraform Enterprise API token to be used with your happy organization
     required: true
+  working_directory:
+    description: "The happy project root"
+    default: "."
+    required: false
```
hspitzley-czi, line 21: I think we also want to add this var to `install-happy` and pass it through in this action so it knows where to find the `.happy/version.lock`

jakeyheath (dismissed):

timzchang (approved):

.github/actions/happy-cleanup/action.yml
shitchcock: I don't understand why we would need to add it to `install-happy` when the input var works fine in [deploy-happy-stack](https://github.com/chanzuckerberg/github-actions/blob/main/.github/actions/deploy-happy-stack/action.yml#L42-L45)? Which uses install happy in a similar fashion

jakeyheath: This is probably fine for now since the edu version of the happy cleanup script has the version hard coded in their action. Generally, we should try to grab the project's version so they match using this input: https://github.com/chanzuckerberg/github-actions/blob/main/.github/actions/install-happy/action.yaml#L20C2-L20C21. But maybe we can cut a new PR for that feature.

jakeyheath (approved):

.github/actions/happy-cleanup/action.yml
jakeyheath: Ref: https://github.com/chanzuckerberg/edu-github-actions/blob/main/.github/actions/clean-happy-rdev/action.yml#L25

shitchcock: We are using deploy without having to declare a version lock file. Should we update our callsites?

https://github.com/chanzuckerberg/bento/blob/main/.github/workflows/deploy.yml#L184-L194

PR #268, opened 2024-05-29
feat: embed path filtering into Argus builder workflow

hspitzley-czi:
Adds arguments to allow users to specify the paths that the action should run on. IE, action will only run on commits that have changed files matching the path filters

### DIFF
--- a/.github/workflows/argus-docker-build.yaml
+++ b/.github/workflows/argus-docker-build.yaml
@@ -23,6 +23,7 @@ on:
           This option is ignored if action is triggered by pull_request event.
         required: false
         type: string
+        default: ${{ github.ref }}
 
 jobs:
   prep:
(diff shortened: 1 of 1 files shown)


.github/workflows/argus-docker-build.yaml
```diff
@@ -11,6 +11,18 @@ on:
...
         required: true
         type: string
+      path_filters:
+        description: 'Path to the configuration file or YAML string with filters definition'
+        required: false
+        type: string
+        default: '**/*'
+      path_filters_base:
+        description: |
+          Git reference (e.g. branch name) against which the changes will be detected. Defaults to repository default branch (e.g. master).
```
jakeyheath, line 21: can the default be {{ github.ref }}?

hspitzley-czi: yep, good call!

jakeyheath (approved):

.github/workflows/argus-docker-build.yaml
```diff
@@ -11,6 +11,19 @@ on:
...
```
jakeyheath, line 21: update description

PR #275, opened 2024-06-27
feat: add ability to delete all stacks in happy environment

shitchcock:
We have a bunch of stacks that don't belong to apps with automated happy cleanup jobs.

We want the ability to delete any stale stacks in the whole edu environment 

https://czi-tech.atlassian.net/jira/software/c/projects/EFI/boards/209?selectedIssue=EFI-1888

Test details. Open up bash and run commands

```
ENV="rdev"
list="happy list --all --output json --env $ENV"
force=""
for i in $($(echo $list) | jq -r --arg date "$date" '.[] | select(.last_updated < $date) | .stack'); do
          echo "Deleting stack: $i"
done
```
The following fails
```
happy delete smp-shitchcock --env $ENV "$force"
```
The following succeeds
```
force="--force"
happy delete smp-shitchcock --env $ENV "$force"
```


timzchang (commented):
Where do we plan to run this action with `all: true`?

Also check for understanding, does this mean running this action in edu-platform-observability with `all: true` will remove all stale stacks for different apps?

shitchcock:
> Where do we plan to run this action with `all: true`?

I plan on writing a Github action that runs once every couple weeks and deletes any stacks in edu platform that have been running for longer than 30 days.

> Also check for understanding, does this mean running this action in edu-platform-observability with `all: true` will remove all stale stacks for different apps?

Yes for all stale stacks in the edu platform happy environment

ahuth (approved):

PR #283, opened 2024-07-11
feat: break argus-docker-build workflow into composite actions

hspitzley-czi:
**Changes:**
- This breaks the `argus-docker-build` Reusable Action into a handful of Composite Actions, making it more composable/customizable. 
- This also adds support for monorepos by allowing users to specify the `working_directory` in case their `.infra` directory is not at the repo root

**Backwards compatibility:** 
The existing Workflow has been refactored to delegate out to the new Composite Actions, meaning we continue to provide a simpler way to build images for Argus when customization is not necessary.

**Reasoning:**
This was done to allow applications (such as Along) to use Github repo/org secrets as Docker build_args.


.github/actions/argus-builder/docker-build/action.yml
```diff
@@ -0,0 +1,98 @@
...
+          IMAGE_TAG=${{ inputs.image_tag }}
+          ${{ inputs.build_args }}
+
+    # TODO: scan image for vulnerabilities
+    # - name: Scan for vulnerabilities
+    #   uses: chanzuckerberg/github-actions/.github/actions/argus-builder/scan-for-vulnerabilities@main
+    #   with:
+    #     image_uri: ${{ steps.ecr_metadata.outputs.ECR_REGISTRY }}/${{ steps.ecr_metadata.outputs.ECR_REPO_NAME }}:${{ inputs.image_tag }}
+    #     github_app_id: ${{ inputs.github_app_id }}
+    #     github_private_key: ${{ inputs.github_private_key }}
```
hspitzley-czi, line 98: Will address in a followup

alexlokshin-czi (approved):

PR #301, opened 2024-08-22
fix: add fallback to force push detection

hspitzley-czi:


alexlokshin-czi (approved):

hspitzley-czi:
Triggering this on pull_request is not a good workflow for a handful of reasons. Primarily:

If the file change detection looks at the whole PR to decide whether a file changed that requires rebuild, it will lead to a build cycle because each commit will retrigger and it will see all the files that changed and build again (this is what we saw start to happen in your PR)

If the file change detection looks at only the most recent commit and the stack label is added to the PR when the most recent commit doesn’t include a chance to a file that would require building the image, it won’t build and I think this will be an confusing user experience

There are also complications with getting the correct commit sha

PR #323, opened 2024-11-01
fix: try supporting workflow_dispatch event type here

kuannie1:


jakeyheath:
is this being worked on @kuannie1 . just trying to keep our PRs clean and not stale

PR #331, opened 2025-01-15
fix: do not update manifest on PRs without stack label

wontonst:
Attempts to address https://github.com/chanzuckerberg/argus/issues/655 by running `Update ArgoCD manifests` only if a PR associated with a push has the `stack` label.

Tested via https://github.com/chanzuckerberg/core-platform-example-app/pull/112 which is on 2.X of this action (hence the merge conflicts).

If we like this approach, what should I do? Maybe create 2.16.0 and 3.7.0? Do we care about supporting 2.X apps? It would be simpler to ask those app owners to use 3.X to gain this benefit. Prior to merging I will rebase on latest and test again on an app using 3.X.

Requires application owners to upgrade their argus-docker-build.yaml version target

- [ ] after merging update the tag of build-prep

### DIFF
--- a/.github/actions/argus-builder/build-prep/action.yml
+++ b/.github/actions/argus-builder/build-prep/action.yml
@@ -21,6 +21,13 @@ inputs:
     description: 'Branch names to run this job on, supports wildcards, comma delimited'
     required: false
     default: ''
+  pr_labels:
+    description: 'Labels attached to an associated PR'
+    required: false
+    default: ''
+  manifest_trigger_label:
+    description: 'Label that will trigger updating the manifest and committing a docker image change to values.yaml'
+    default: 'stack'
 
 outputs:
   image_tag:
@@ -29,6 +36,9 @@ outputs:
   should_build:
     description: Whether the job should run
     value: ${{ steps.final_check.outputs.should_build }}
+  should_deploy:
+    description: Whether or not the build is to be deployed to a stack
+    value: ${{ steps.require_stack_label.outputs.should_deploy }}
 
 runs:
   using: composite
@@ -94,6 +104,7 @@ runs:
           const filters = `${{ inputs.path_filters }}`.split(',').map(f => f.trim()).filter(b => b.length > 0);
           const filtersStr = "run_on:\n" + filters.map(f => `  - '${f}'`).join('\n');
           core.setOutput('filters', filtersStr);
+          console.log(`Filters: ${filtersStr}`);
 
     - name: Check for force push
       id: force_push
@@ -119,6 +130,17 @@ runs:
         base: ${{ steps.force_push.outputs.base }}
         list-files: json
 
+    - name: Require manifest trigger label in pull requests
+      id: require_stack_label
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const labels = `${{ inputs.pr_labels }}`;
+          const triggerLabel = `${{ inputs.manifest_trigger_label }}`;
+          const hasTriggerLabel = labels.split(',').map(l => l.trim()).includes(triggerLabel);
+          console.log(`Detected labels ${labels} on PR. Has ${triggerLabel} label? ${hasTriggerLabel}`);
+          core.setOutput('should_deploy', hasTriggerLabel);
+
     - name: Check if build should run
       id: final_check
       uses: actions/github-script@v7
@@ -126,4 +148,6 @@ runs:
         script: |
           const branchMatched = `${{ steps.branch_filter.outputs.match }}` === 'true';
           const filesMatched = `${{ steps.file_filter.outputs.run_on }}` === 'true';
-          core.setOutput('should_build', filesMatched && branchMatched);
+          const shouldBuild = filesMatched && branchMatched;
+          core.setOutput('should_build', shouldBuild);
+          console.log(`Branch matched? ${branchMatched}. Files matched? ${filesMatched}. Build should run? ${shouldBuild}`);
--- a/.github/workflows/argus-docker-build.yaml
+++ b/.github/workflows/argus-docker-build.yaml
@@ -48,20 +48,26 @@ jobs:
     outputs:
       image_tag: ${{ steps.build_prep.outputs.image_tag }}
       should_build: ${{ steps.build_prep.outputs.should_build }}
+      should_deploy: ${{ steps.build_prep.outputs.should_deploy }}
       images: ${{ steps.parse_images.outputs.images }}
     permissions:
       id-token: write
       contents: read
       pull-requests: read
     steps:
-      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@v3.4.1
+      - uses: 8BitJonny/gh-get-current-pr@3.0.0
+        id: findPr
+        with:
+          filterOutClosed: true
+          filterOutDraft: false
+      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@rzheng/stack-label
         id: build_prep
         with:
           path_filters: ${{ inputs.path_filters }}
           path_filters_base: ${{ inputs.path_filters_base }}
           branches_include: ${{ inputs.branches_include }}
           branches_ignore: ${{ inputs.branches_ignore }}
-      
+          pr_labels: ${{ steps.findPr.outputs.pr_labels }}
       - uses: actions/checkout@v4
       - uses: chanzuckerberg/github-actions/.github/actions/validate-json-schema@v3.4.1
         name: Validate images input
@@ -219,7 +225,7 @@ jobs:
     permissions:
       id-token: write
       contents: read
-    if: always() && inputs.update_manifests && needs.prep.outputs.should_build == 'true' && needs.prep.outputs.images != '[]'
+    if: always() && inputs.update_manifests && needs.prep.outputs.should_build == 'true'&& needs.prep.outputs.should_deploy == 'true' && needs.prep.outputs.images != '[]'
     steps:
       - name: Check build status
         uses: actions/github-script@v7


.github/workflows/argus-docker-build.yaml
```diff
@@ -52,14 +52,21 @@ jobs:
...
       contents: read
+      pull-requests: read
     steps:
-      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@c210eb8dba1aeb3ccf02186ad89c3c85f6fc4da7
+      - uses: 8BitJonny/gh-get-current-pr@3.0.0
+        id: findPr
+        with:
+          filterOutClosed: true
+          filterOutDraft: false
+      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@rzheng/stack-label
```
wontonst, line 62: what is our versioning strategy? continue using a sha1? leave it pointing to a branch like this PR? maybe we can create a tag?

.github/actions/argus-builder/build-prep/action.yml
```diff
@@ -120,4 +134,5 @@ runs:
         script: |
           const branchMatched = `${{ steps.branch_filter.outputs.match }}` === 'true';
           const filesMatched = `${{ steps.file_filter.outputs.run_on }}` === 'true';
-          core.setOutput('should_build', filesMatched && branchMatched);
+          const prHasStackLabel = `${{ steps.require_stack_label.outputs.has_label }}` === 'true';
+          core.setOutput('should_build', filesMatched && branchMatched && prHasStackLabel);
```
wontonst, line 138: note this also skips `Update Manifest`

is that okay or should I modify the manifest-update action to only skip the step that adds the values.yaml change?

.github/workflows/argus-docker-build.yaml
hspitzley-czi: we use `release-please` to propose releases based on our conventional commits and merging the release proposal PR creates the actual release/tag. the reason this uses a sha instead of one of those tags is that it allows us to make a single release with the update instead of first releasing `build-prep` and then coming back and updating the version of `build-prep` used in this file and making another release

.github/workflows/argus-docker-build.yaml
```diff
@@ -52,14 +52,21 @@ jobs:
     permissions:
       id-token: write
       contents: read
+      pull-requests: read
     steps:
-      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@c210eb8dba1aeb3ccf02186ad89c3c85f6fc4da7
+      - uses: 8BitJonny/gh-get-current-pr@3.0.0
```
hspitzley-czi, line 57: does this work on `push` and on `pull_request`?

jakeyheath: From the docs: 

`Inside a running github action you can't always get the information whether you are currently in a PR or not. If for example the trigger event is pull_request you have that information but not when your trigger event is push.`

.github/actions/argus-builder/build-prep/action.yml
```diff
@@ -21,6 +21,10 @@ inputs:
     description: 'Branch names to run this job on, supports wildcards, comma delimited'
     required: false
     default: ''
+  labels:
+    description: 'Labels attached to an associated PR'
+    requred: false
+    default: ''
```
jakeyheath, line 27: rename to something that indicates these are the labels currently attached to the PR. Then we need another input for the name of the label with the default value being "stack". I think we should allow for the name of the label to be configured, with a sane default

.github/actions/argus-builder/build-prep/action.yml
```diff
@@ -104,6 +108,16 @@ runs:
             core.setOutput('base', `${{ inputs.path_filters_base }}`);
           }
 
+    - name: Require stack label in pull requests
+      id: require_stack_label
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const labels = `${{ inputs.labels }}`;
+          let hasStackLabel = labels.split(',').includes("stack")
```
jakeyheath, line 117:
```suggestion
          let hasStackLabel = labels.split(',').includes("{{inputs.label}}")
```

see comment above

.github/actions/argus-builder/build-prep/action.yml
wontonst: done

.github/actions/argus-builder/build-prep/action.yml
wontonst: done

.github/actions/argus-builder/build-prep/action.yml
wontonst: oh wait that step only writes the values.yml changes so it's safe to skip

wontonst:
PR should be ready for review now.

- Rebased this PR latest 3.X
- Upgraded core sample app to 3.6.0
- Tested this change in https://github.com/chanzuckerberg/core-platform-example-app/pull/112/files

Todo:
- [ ] Pin argus-builder/build-prep to a commit that is approved by reviewers
- [ ] Bump argus-docker-build version in relevant repos

Side note I noticed that rerunning jobs via the Github UI when a values.yaml commit has been created always results in the manifest update step not getting executed. My guess is because the file filters thinks the base commit is the commit before the values.yaml file commit so it thinks there are no files matching the path filters. I think we can ignore this since most folks are probably pushing to their branch to trigger a build but good to know since I wasted time figuring out why steps were getting skipped.

.github/actions/argus-builder/build-prep/action.yml
```diff
@@ -110,6 +118,17 @@ runs:
...
           }
 
+    - name: Require manifest trigger label in pull requests
+      id: require_stack_label
+      uses: actions/github-script@v7
+      with:
+        script: |
+          const labels = `${{ inputs.pr_labels }}`;
+          const triggerLabel = `${{ inputs.manifest_trigger_label }}`;
+          let hasTriggerLabel = labels.split(',').includes(triggerLabel)
```
hspitzley-czi, line 128: Some apps need to call this directly instead of through the reusable workflow so let's add a `trim` to sanitize their input
```suggestion
          const hasTriggerLabel = labels.split(',').map(l => l.trim()).includes(triggerLabel)
```

.github/actions/argus-builder/build-prep/action.yml
```diff
@@ -126,4 +145,7 @@ runs:
         script: |
           const branchMatched = `${{ steps.branch_filter.outputs.match }}` === 'true';
           const filesMatched = `${{ steps.file_filter.outputs.run_on }}` === 'true';
-          core.setOutput('should_build', filesMatched && branchMatched);
+          const prHasStackLabel = `${{ steps.require_stack_label.outputs.has_trigger_label }}` === 'true';
+          const shouldBuild = filesMatched && branchMatched && prHasStackLabel;
+          core.setOutput('should_build', shouldBuild);
```
hspitzley-czi, line 150: it might be good to have the image still get built, the benefits of that would be even without making a stack a user would still be able to see whether their code introduced a bug that prevents the image from building. @jakeyheath what do you think?

.github/actions/argus-builder/build-prep/action.yml
wontonst: done

.github/actions/argus-builder/build-prep/action.yml
wontonst: I think given that the main artifact for a build is often times an image, and given the low cost of building an image, we should just always build an image. what do we think about removing `should_build` from  build-docker's `if` condition?

```
if: needs.prep.outputs.should_build == 'true' && needs.prep.outputs.images != '[]'
```

alternatively I can create a new output like `shouldDeploy` and use that to separately control `Update manifest`

.github/actions/argus-builder/build-prep/action.yml
```diff
@@ -21,6 +21,13 @@ inputs:
     description: 'Branch names to run this job on, supports wildcards, comma delimited'
     required: false
     default: ''
+  pr_labels:
+    description: 'Labels attached to an associated PR'
+    required: false
+    default: ''
+  manifest_trigger_label:
```
hspitzley-czi, line 28: I think this needs to allow multiple values, comma delimited

We don't have anyone with this setup right now but we've previously had people build a monorepo with multiple Argus apps that use independent labels to control stack creation. We have an example monorepo set up to work like that: https://github.com/chanzuckerberg/argus-example-monorepo

wontonst: hmm, this could get confusing. some folks might want it to mean "any of these labels should trigger a manifest update" and others "these labels must all be applied to trigger a manifest update"

what do you think about crossing that bridge when we come to it?

hspitzley-czi: I'm good with that, alternatively it could take a regex which would make it support the "any of these labels" case

.github/actions/argus-builder/build-prep/action.yml
jakeyheath: yea that would be more clear @wontonst . Let's go forward with that

.github/workflows/argus-docker-build.yaml
```diff
@@ -54,14 +54,19 @@ jobs:
       contents: read
       pull-requests: read
     steps:
-      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@v3.4.1
+      - uses: 8BitJonny/gh-get-current-pr@3.0.0
+        id: findPr
+        with:
+          filterOutClosed: true
+          filterOutDraft: false
+      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@rzheng/stack-label
```
jakeyheath, line 62: Don't forget to change this in a follow up PR

jakeyheath (approved):

.github/workflows/argus-docker-build.yaml
wontonst: https://github.com/chanzuckerberg/github-actions/pull/335

wontonst:
will defer upgrading other repos until https://czi.atlassian.net/browse/CCIE-3947

PR #333, opened 2025-01-22
feat!: CCIE-3896 rename CZI_GITHUB_HELPER_* to GH_ACTIONS_HELPER_*

wontonst:
BREAKING CHANGE: requires renaming of org level secrets

* Might consider bundling together with other breaking changes (if any) that may arise from making this repo reusable across orgs
* Cannot rename to `GITHUB_HELPER_*` bc the `GITHUB_` prefix is reserved
* I don't think the GH app is checked into IaC right? anyone reading the README will likely have to reach out to our team to figure out app settings/permissions.
* I generated a new PK for the `CZI Github Helper` app and added `GH_ACTIONS_HELPER_APP_ID` & `GH_ACTIONS_HELPER_PK` as new org secrets. we should delete the old ones after this merges.

### DIFF
--- a/.github/workflows/argus-docker-build.yaml
+++ b/.github/workflows/argus-docker-build.yaml
@@ -215,8 +215,8 @@ jobs:
           build_args: ${{ matrix.image.build_args }}
           secret_files: ${{ matrix.image.secret_files }}
           image_tag: ${{ needs.prep.outputs.image_tag }}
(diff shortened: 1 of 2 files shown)


jakeyheath (approved):

wontonst:
Merge this after https://github.com/chanzuckerberg/github-actions/pull/331 lands so people can get that change w/o having to upgrade major version

PR #335, opened 2025-01-24
chore: CCIE-3899 use build-prep@3.6.1

wontonst:
3.6.1 release pr merged https://github.com/chanzuckerberg/github-actions/pull/334

### DIFF
--- a/.github/workflows/argus-docker-build.yaml
+++ b/.github/workflows/argus-docker-build.yaml
@@ -60,7 +60,7 @@ jobs:
         with:
           filterOutClosed: true
           filterOutDraft: false
-      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@rzheng/stack-label
+      - uses: chanzuckerberg/github-actions/.github/actions/argus-builder/build-prep@3.6.1
         id: build_prep
         with:
           path_filters: ${{ inputs.path_filters }}


hspitzley-czi (approved):

wontonst:
how do we feel about unpinning the version?

wontonst:
also, I think it might make sense to create multiple tags per release, eg

3.6.1
3.6-latest
3-latest

one drawback of the current setup is it's not clear why a major bump occurred since there are so many actions in this repo.

ways to overcome

- break the repo into one action per repo. I don't like this since it makes finding the action you want harder
- add the action name to the tag eg build-prep-3.6.1 but the automation to do this would be pretty complex and `git tag -l` would have a lot of clutter. I think someone tried this already since I see `latest-successful-deployment-v1.0.0`. doesn't seem worth it.
- culturally optimize against major version bump, unpin everything. if a major bump is needed then the person doing the bump must migrate all consumers of the action. forces us to think harder about action design but maybe that's not a bad thing.

hspitzley-czi:
> add the action name to the tag eg build-prep-3.6.1 but the automation to do this would be pretty complex and git tag -l would have a lot of clutter. I think someone tried this already since I see latest-successful-deployment-v1.0.0. doesn't seem worth it.

this is actually pretty easy with release-please. the directories in the `actions` subdir can be made to get their own tags easily, i'm not sure how it would work for things in `workflows` since it just contains files, I haven't tried that

PR #340, opened 2025-01-28
chore: CCIE-3899 use sha to fetch PR in argus-docker-build

wontonst:


.github/workflows/argus-docker-build.yaml
```diff
@@ -60,6 +60,7 @@ jobs:
         with:
           filterOutClosed: true
           filterOutDraft: false
+          sha: ${{ github.event.pull_request.head.sha }}
```
hspitzley-czi, line 63: If this is triggered by a push `github.event.pull_request.head.sha` won't have a value, right? Will that cause any issues?

wontonst: I tested it with a push and it still works

hspitzley-czi (approved):

PR #343, opened 2025-02-03
chore: CCIE-3986 use persist-credentials: false in checkout

wontonst:
I did a manual inspection of the files but no guarantees that this won't break something.

If anyone knows of a place where we run privileged git commands other than update manifest, please let me know.

If this PR doesn't end up breaking anything I'll open PRs for other repos using the same approach.

### DIFF
--- a/.github/actions/argus-builder/manifest-update/action.yml
+++ b/.github/actions/argus-builder/manifest-update/action.yml
@@ -65,6 +65,7 @@ runs:
           }
     - uses: actions/checkout@v4
       with:
+        persist-credentials: true # creds used in Update Argus Manifest step
         fetch-depth: 0
         token: ${{ steps.generate_token.outputs.token }}
         ref: ${{ steps.ref.outputs.result }}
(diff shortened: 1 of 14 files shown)


.github/actions/argus-builder/manifest-update/action.yml
```diff
@@ -51,6 +51,7 @@ runs:
           }
     - uses: actions/checkout@v4
       with:
+        persist-credentials: true # creds used in Update Argus Manifest step
```
wontonst, line 54: not sure if we just accept this risk or if we want to do something about this

in my experience, automation doesn't make commits against the repo. the name of the artifact is also not something baked into the repo that get constantly updated. it's usually an external system managing the mapping of git hash to docker img tag, and that system will also keep track of what's deployed in prd/staging. sounds like that should be all done in argus so I'm curious if that's something that's possible/something we've considered doing.

.github/workflows/argus-docker-build.yaml
```diff
@@ -110,6 +110,7 @@ jobs:
         image: ${{ fromJson(needs.prep.outputs.images) }}
     steps:
       - uses: actions/checkout@v4
+        persist-credentials: false
```
latkinsa, line 113: ` with:` is missing here; my VSCode is very unhappy about it 😄

PR #345, opened 2025-02-04
chore: CCIE-3978 remove Linux & self-hosted github runner tags

wontonst:


hspitzley-czi (changes requested):
Looks good but I think the self-hosted runners still need a little work before we can merge

jakeyheath:
These labels will work with our old runner too, so probably safe.

jakeyheath (approved):

wontonst:
@hspitzley-czi when do you think this PR will become mergeable?

hspitzley-czi:
> These labels will work with our old runner too, so probably safe.

This is true but if we iterate on the new runners and release a version of them that can't execute jobs then fixing them will become urgent rather vs just not sending jobs to the new runners until they are stable

> @hspitzley-czi when do you think this PR will become mergeable?

Deferring to @jakeyheath

hspitzley-czi (approved):
Think the new runners are good to go now

wontonst:
well that was fast 😄

PR #349, opened 2025-02-06
feat: CCIE-3978 make argus-docker-build runs-on configurable

wontonst:
Considerations:

1. Github workflows apparently do not allow using a list as an input. I want to support the runs-on labels with more than 1 selector so a workaround I found is to pass a JSON string that gets decoded using `fromJSON()`.
2. Github-hosted arm-based runners are not enabled for private repos
image
3. There is no way to use x64 for both build-prep and x64 matrix img build steps (not sure if there's a use case for this)

tested in https://github.com/chanzuckerberg/core-platform-example-app/pull/112

are there any other workflows that we want to make `runs-on` configurable? `install-happy` and `test-tfe-cache` are using custom runners but not sure if those should be configurable

### DIFF
--- a/.github/workflows/argus-docker-build.yaml
+++ b/.github/workflows/argus-docker-build.yaml
@@ -44,11 +44,21 @@ on:
         required: false
         type: boolean
         default: false
+      arm_runner_labels:
+        description: 'Runner label to use for steps that use an arm-based action runner'
+        required: false
+        type: string
+        default: "['ARM64']"
+      x64_runner_labels:
+        description: 'Runner label to use for steps that use an x64-based action runner'
+        required: false
+        type: string
+        default: "['X64']"
 
 jobs:
   prep:
     name: Prep for Build
-    runs-on: [ARM64]
+    runs-on: ${{ fromJSON(inputs.arm_runner_labels) }}
     if: contains(github.event.head_commit.message, '[no-deploy]') == false
     outputs:
       image_tag: ${{ steps.build_prep.outputs.image_tag }}
@@ -191,8 +201,7 @@ jobs:
   build-docker:
     name: Build Docker Image
     needs: [prep]
-    runs-on:
-      - ${{ matrix.image.platform == 'linux/amd64' && 'X64' || 'ARM64' }}
+    runs-on: ${{ matrix.image.platform == 'linux/amd64' && fromJSON(inputs.x64_runner_labels) || fromJSON(inputs.arm_runner_labels) }}
     if: needs.prep.outputs.should_build == 'true' && needs.prep.outputs.images != '[]'
     permissions:
       id-token: write
@@ -225,7 +234,7 @@ jobs:
   update-manifests:
     name: Update ArgoCD manifests
     needs: [prep, build-docker]
-    runs-on: [ARM64]
+    runs-on: ${{ fromJSON(inputs.arm_runner_labels) }}
     permissions:
       id-token: write
       contents: read


.github/workflows/argus-docker-build.yaml
```diff
@@ -44,11 +44,21 @@ on:
         required: false
         type: boolean
         default: false
+      arm_runner:
```
jakeyheath, line 47:
```suggestion
      arm_runner_labels:
```

jakeyheath: In case we want to support other labels 🤷🏻

.github/workflows/argus-docker-build.yaml
```diff
@@ -44,11 +44,21 @@ on:
...
+        description: 'Runner label to use for steps that use an arm-based action runner'
+        required: false
+        type: string
+        default: "['ARM64']"
+      x64_runner:
```
jakeyheath, line 52:
```suggestion
      x64_runner_labels:
```

jakeyheath:
> Github workflows apparently do not allow using a list as an input. I want to support the runs-on labels with more than 1 selector so a workaround I found is to pass a JSON string that gets decoded using fromJSON().

Ya its annoying lol. @hspitzley-czi and I have run into this before

jakeyheath:
> are there any other workflows that we want to make runs-on configurable? install-happy and test-tfe-cache are using custom runners but not sure if those should be configurable

there might be, but for now don't worry about those other actions. They are 1.) old and 2.) unused. Test-tfe-cache I think was my experiment to try to do terraform applies in GHA instead of TFE. For now, we'll test this out and if the patterns works well, we can adopt it on others.

jakeyheath:
Number 2 and 3 considerations might need more explanation. I'm not sure I understand the scenarios

jakeyheath (approved):

.github/workflows/argus-docker-build.yaml
wontonst: done

wontonst:
> Number 2 and 3 considerations might need more explanation. I'm not sure I understand the scenarios

2. As a developer in a GH org w/ no self hosted runners I would like to run my build on an ARM machine.
3. As a developer I would like to run the `prep` step on x64 which means with this PR I would need to pass `arm_runner_labels=[x64 machine label]`. This is problematic because I still want the arm runner to run on arm. This could be resolved by adding a `build_prep_runner` input but I'm not sure if there's a use case for this.

.github/workflows/release.yaml
```diff
@@ -21,3 +21,16 @@ jobs:
...
+        git config user.name github-actions[bot]
+        git tag -d v${{ steps.release.outputs.major }} || true
+        git tag -d v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }} || true
+        git push origin :v${{ steps.release.outputs.major }} || true
+        git push origin :v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }} || true
+        git tag -a v${{ steps.release.outputs.major }} -m "Release v${{ steps.release.outputs.major }}"
+        git tag -a v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }} -m "Release v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }}"
+        git push origin v${{ steps.release.outputs.major }}
+        git push origin v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }}
+      if: ${{ steps.release.outputs.release_created }}
```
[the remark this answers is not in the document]
wontonst: oh this is totally not meant to be checked in. removed.

hspitzley-czi (approved):

jakeyheath (approved):

PR #351, opened 2025-02-14
feat: don't use github actions cache; we use pull through cache

jakeyheath:
## Summary

Our ECR pull through cache will be faster for self-hosted runners rather than storing on Github. This isn't needed.

## References

* https://github.com/docker/setup-qemu-action/issues/166
* https://github.com/docker/setup-qemu-action/pull/130


wontonst (approved):
curious - how much faster is are things before and after this PR?

are cache objects not shared across workflow runs? I thought the ask in https://github.com/docker/setup-qemu-action/issues/166 is to eliminate unnecessary `docker pull`s

jakeyheath:
> curious - how much faster is are things before and after this PR?
> 
> are cache objects not shared across workflow runs? I thought the ask in [docker/setup-qemu-action#166](https://github.com/docker/setup-qemu-action/issues/166) is to eliminate unnecessary `docker pull`s

Not sure. But the ECR location is in the same network as the self-hosted runner, so I would guess its faster to go in network to pull this image. Since we are using a pull-through ECR cache already, its just an extra API call to also store in the Github cache.

PR #353, opened 2025-02-19
feat!: *BREAKING* allow passing multiple labels that should make argus builder update manifests

hspitzley-czi:
This is a breaking change because it removes the `manifest_trigger_label` arg, replacing it with the pluralized `manifest_trigger_labels`. Impact will be minimal because I don't think anyone uses the old arg.

Tested in https://github.com/chanzuckerberg/metahub/pull/106


wontonst (approved):
nice 👍

PR #358, opened 2025-02-24
feat: CCIE-3947 tag unpinned major/minor versions

wontonst:


.github/workflows/release.yaml
```diff
@@ -21,3 +21,17 @@ jobs:
...
+          git config user.email 95879977+czi-github-helper[bot]@users.noreply.github.com
+          git tag -d v${{ steps.release.outputs.major }} || true
+          git tag -d v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }} || true
+          git push origin :v${{ steps.release.outputs.major }} || true
+          git push origin :v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }} || true
+          git tag -a v${{ steps.release.outputs.major }} -m "Release v${{ steps.release.outputs.major }}"
+          git tag -a v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }} -m "Release v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }}"
+          git push origin v${{ steps.release.outputs.major }}
+          git push origin v${{ steps.release.outputs.major }}.${{ steps.release.outputs.minor }}
+        if: ${{ steps.release.outputs.release_created }}
```
hspitzley-czi, line 37: This is great, what do you think about putting it in a composite action (maybe with the steps above it or maybe not, I'm not really sure)? I ask because I know @jakeyheath wants to do something similar for our Terraform modules

wontonst: I like this idea. the TF PR is at https://github.com/chanzuckerberg/shared-infra/pull/10726 and I think you're right it makes sense to consolidate

jakeyheath: this is going to be great, I agree having with having a consolidated action

wontonst: done.

instead of spending time getting massaging this to be testable I'm thinking I merge this in and quickly fix any issues that come up.  sounds good?

jakeyheath: I wouldn't have it any other way

.github/actions/release-please-semvar/action.yml
```diff
@@ -0,0 +1,32 @@
+name: Release please
+description: Runs release please and tags unpinned major and minor versions
+runs:
+  using: "composite"
+  steps:
+    # See https://docs.github.com/en/actions/security-guides/automatic-token-authentication#using-the-github_token-in-a-workflow
+    # For why we need to generate a token and not use the default
+    - name: Generate token
```
jakeyheath, line 8: did you want to use that action supported by Github here for making a token>

jakeyheath (approved):

.github/actions/release-please-semvar/action.yml
wontonst: done

PR #360, opened 2025-02-26
fix: CCIE-3947 unpin release please git rev in release.yaml

wontonst:


wontonst:
merging without approval to fix release please. this PR does not seem dangerous.

PR #386, opened 2025-03-20
feat!: convert argus-builder build-prep action to TS

hspitzley-czi:
This PR replaces the composite action at `argus-builder/build-prep/` with one written in Typescript

Why is this necessary?
1. It removes dependencies on 3rd party actions - Making it more secure
2. It removes the need to run `npm install <pkg>` at runtime - This causes problems for TS/JS apps because when you run `npm install <pkg>` it tries to install `<pkg>` plus **everything** in package.json. This fails when the package.json uses a package from a private registry because the npmrc file is not present
3. It makes it possible to write unit tests - Providing more assurance that changes don't break things
4. It makes the code more reusable - The "overall" checks we do to determine if *any* images should be built are essentially the same as the per-image checks we do (albeit with different inputs). In the YAML version that logic was re-implemented but the TS version is better able to re-use the same functions, providing more consistency and easier maintainability

Tested in:
- argus-example-app: https://github.com/chanzuckerberg/argus-example-app/pull/232
- argus-example-monorepo: https://github.com/chanzuckerberg/argus-example-monorepo/pull/27


.github/actions/argus-builder/build-prep/action.yml
```diff
@@ -1,18 +1,17 @@
...
+    description: The GitHub token for authenticating with the GitHub API
+    required: true
+  images:
+    description: 'JSON object specifying the images to build'
+    required: true
   path_filters:
     description: 'Glob patterns to match against changed files in the repository, comma delimited'
     required: false
     default: '**/*'
-  path_filters_base:
```
wontonst, line 9: I think we'll need to do a major version bump to remove this

.github/actions/argus-builder/build-prep/src/main.ts
```diff
@@ -0,0 +1,259 @@
...
+    githubToken: core.getInput('github_token', { required: true }),
+    images: JSON.parse(core.getInput('images', { required: true })),
+    pathFilters: getCommaDelimitedArrayInput('path_filters', { required: true }),
+    branchesInclude: getCommaDelimitedArrayInput('branches_include', { required: true }),
+    branchesIgnore: getCommaDelimitedArrayInput('branches_ignore', { required: false }),
+    manifestTriggerLabels: getCommaDelimitedArrayInput('manifest_trigger_labels', { required: true }),
+  };
+}
+
+const imagesInputSchema = {
```
wontonst, line 34: this is new right? I think we'll need to document this somewhere with examples

.github/actions/argus-builder/build-prep/src/main.ts
```diff
@@ -0,0 +1,259 @@
...
+
+if (process.env.NODE_ENV !== 'test') {
+  main();
+}
+export async function main() {
+  const inputs = getInputs();
+  core.info(`Received input: ${JSON.stringify(inputs, null, 2)}`);
+
+  core.info('Validating images input...');
+  validateJsonSchema(inputs.images, imagesInputSchema);
```
wontonst, line 88: I'm not familiar with Ajv - inputs.images is an array right? imagesInputSchema says the root is of type object  but shouldn't it be an array?

wontonst:
how is the combined dist js file created? ncc? would be good to document this somewhere, and maybe eventually add automation

.github/workflows/argus-docker-build.yaml
```diff
@@ -65,15 +67,13 @@ on:
           If ANY of the labels are found on the PR, the manifest will be updated.
         required: false
         type: string
-        default: 'stack'
+        default: 'stack,stack-*'
```
wontonst, line 70: did someone request stack-*?

.github/actions/argus-builder/build-prep/src/main.ts
hspitzley-czi: Sort of, it was formerly in `argus-docker-build.yaml`: https://github.com/chanzuckerberg/github-actions/blob/85706b317c52de7c9b4b7249d15d81f41bc5ac2e/.github/workflows/argus-docker-build.yaml#L101-L145

.github/actions/argus-builder/build-prep/src/main.ts
hspitzley-czi: `inputs.images` in an object, the `processImagesInput` function converts it to an array which is written to the output and that output is what the build step uses to set up the build matrix

.github/workflows/argus-docker-build.yaml
hspitzley-czi: not `stack-*` explicitly but we've seen the metahub project want to use `stack-science` and edu point solutions needing `stack-notifications`

I'm not attached to this change but thought it might be nice to not have to specify this input in most cases

hspitzley-czi:
> how is the combined dist js file created? ncc? would be good to document this somewhere, and maybe eventually add automation

yep, `ncc` converts the TS to native JS that the runners can invoke. Automated compile, lint, and test jobs are already running and committing changes to dist back to PRs, and I have a follow-up task in this sprint to write docs on how to make an action like this

.github/actions/argus-builder/build-prep/action.yml
hspitzley-czi: New plan, doing a major bump so I can clean this up since a major bump is going to be needed for the next changes to the `manifest-update` step

.github/actions/argus-builder/build-prep/src/main.test.ts
```diff
@@ -0,0 +1,331 @@
+import {
+  findMatchingChangedFiles, isLabelOnPullRequest, isMatchingBranch, ProcessedImage, processImagesInput, wildcardMatch,
+} from './main';
+
+// mock core to suppress logs
+jest.mock('@actions/core');
+
+describe('validate-json-schema', () => {
+  describe('wildcardMatch', () => {
```
wontonst, line 9: I wonder if anyone uses special characters in their label. will labels that use regex symbols trip up the function? eg `$myvariable` or `build[amd64]`

.github/actions/find-changed-files/src/main.ts
```diff
@@ -19,46 +15,12 @@ export async function main() {
...
   core.info(`Received input: ${JSON.stringify(inputs, null, 2)}`);
 
-  const res = await findChangedFiles(inputs);
+  const res = await findChangedFiles(inputs.githubToken);
 
   core.info(`Result: ${JSON.stringify(res, null, 2)}`);
   core.setOutput('all_modified_files', res.allModifiedFiles.join(','));
 }
 
-export async function findChangedFiles(inputs: Inputs): Promise<Output> {
```
wontonst, line 28: can this be used to replace tj-actions?

wontonst (approved):

.github/actions/find-changed-files/src/main.ts
hspitzley-czi: Partially, it currently only reimplements some of the functionality from `tj-actions-changed-files`, but we can pretty easily extend it to cover more of that functionality as needed

.github/actions/argus-builder/build-prep/src/main.test.ts
hspitzley-czi: Made it work with special characters in [5ef9f21](https://github.com/chanzuckerberg/github-actions/pull/386/commits/5ef9f210f6cd8849e7e8ea63556e1f50536468df) and added tests there to show that it works

hspitzley-czi: Good callout

.github/actions/find-changed-files/src/main.ts
hspitzley-czi: Specifically, it only gives the equivalent of this output: https://github.com/tj-actions/changed-files/blob/27ae6b33eaed7bf87272fdeb9f1c54f9facc9d99/action.yml#L301

Issue #548, opened 2026-06-11
Proposal: Apply-on-PR Terragrunt CI (reusable workflow to replace Terraform Enterprise)


jakeyheath:
# Apply-on-PR Terragrunt CI

### Problem

We want to migrate our automated terraform plans, applies and state syncs from TFE to something else. The contract is being difficult to negociate after their merge with IBM. We have some prior art in this space via `argus-infra-stacks`. It runs its Terraform through a homegrown Terragrunt workflow on GitHub Actions. That workflow has been tested a good amount, and we've seen good successes with it. It's not as fully featured as TFE, but most of our feature requirements are limited. We'd like to generalize this workflow for others to use. 

As a part of this migration, we want to solve a problem many people have with our VCS workflows in TFE. Today, a push to a PR runs a plan, and the apply only runs after the PR merges to `main`. Applies fail often — transient errors, or small mistakes a plan does not catch. When that happens you have already passed CI, gotten review, and merged, so you start the whole cycle over to fix a one-line problem. We would rather catch those failures while the PR is still open. 

### Goal

Generalize the `argus-infra-stacks` workflow into a reusable, repo-agnostic system that any org infrastructure repo can adopt to replace Terraform Enterprise, and move the apply from "after merge to `main`" to "on the PR, before merge." You review the plan, run the apply while the PR is still open, and only merge once the apply has actually succeeded. Failures surface during review instead of after it.

### Approach in brief

Applies are triggered by an `/apply` comment on the PR and run against the post-merge code. A successful apply unblocks an auto-merge, so the branch and `main` converge immediately. Moving applies onto PRs creates three new risks, each handled deliberately:

- **Two PRs clobbering the same stack**: a shared per-stack authority record tracks which PR currently owns a stack. Other PRs are blocked until it is released.
- **The branch diverging from `main`**: both plan and apply run against GitHub's speculative merge commit, and the PR auto-merges right after a clean apply.
- **The nightly drift job reverting unmerged work**: the nightly apply reads the authority record first and skips any stack a PR currently owns.

As a safety net, a post-merge apply still runs against `main` over the merged stacks. Normally it is a no-op (the PR already applied), but it automatically recovers the case where branch protection was weakened or toggled off and a PR merged without applying.

The system stays fogg-agnostic, uses standard GitOps mechanics (no GitHub Environments), and runs under a single org-wide CI role. Moving existing state off Terraform Enterprise is a separate prerequisite (see Scope boundary).

[...]

PR #549, opened 2026-06-12
feat: port terragrunt CI composite actions from argus-infra-stacks

jakeyheath:
## Overview

We are generalizing the `argus-infra-stacks` Terragrunt CI into a reusable, repo-agnostic system that other org repos can adopt to replace Terraform Enterprise (proposal: #548). The first step is moving the composite actions that the central engine depends on into this repository, where the reusable workflow will live.

## Solution

Port three composite actions out of `argus-infra-stacks` and into this actions monorepo:

- `install-aws-cli`
- `release-terraform-locks`
- `terragrunt-provider-cache`

Runtime behavior is unchanged. The bundled `dist/` is rebuilt from the same source as the proven argus actions. Each action is conformed to this repo's conventions:

- Scoped npm workspace names (`@chanzuckerberg/<name>-action`) with `build` / `lint` / `lint:fix` / `test` scripts.
- Tests moved from vitest to jest (the repo standard).
- The shared root `tsconfig.json` and `.eslintrc.js` are used instead of per-action configs.
- The recursive backend-file and cache-dir walkers keep their generator and `for...of` style via a small per-action eslint override, rather than rewriting working logic.

## Impact

No behavior change to any existing workflow. Nothing in this repo references these actions yet. This PR is the bottom of a stack: **#550 is stacked on top of it** and adds the reusable `terragrunt-engine.yml` workflow that consumes these actions. Merge this first.

One follow-up to call out: the `install-aws-cli` `run()` orchestration test was dropped because it relied on vitest's ESM mocking of `@actions/cache` (v6 is ESM-only), which jest cannot resolve under the repo's CJS setup. The pure-function coverage (`lib.test.ts`) is retained, and re-adding the orchestration test under jest is a follow-up.

Part of #548 (PR1 of the delivery plan).


.github/actions/terragrunt-provider-cache/package.json
```diff
@@ -0,0 +1,15 @@
...
+  "version": "0.0.1",
+  "private": true,
+  "scripts": {
+    "test": "jest --config=../../../jest.config.js --rootDir=./",
+    "build": "ncc build src/main.ts --source-map --minify",
+    "lint": "eslint --max-warnings=0 src/",
+    "lint:fix": "eslint --fix src/"
+  },
+  "dependencies": {
+    "@actions/core": "^1.11.1",
```
hspitzley-czi, line 12: How do we want to handle common deps like this? So far I've been putting them in the root-level package.json (`"@actions/core": "^1.11.1"` exists there) but we could shift to making it explicit in every action like this

.github/actions/terragrunt-provider-cache/src/main.ts
```diff
@@ -0,0 +1,6 @@
+import * as core from '@actions/core';
+import { run } from './cache-ops';
+
+run().catch((err) => {
+  core.setFailed(err instanceof Error ? err.message : String(err));
+});
```
hspitzley-czi, line 6: Could be useful to set it up like this so you can have unit tests https://github.com/chanzuckerberg/github-actions/blob/main/.github/actions/find-changed-files/src/main.ts#L26-L28

hspitzley-czi: Er, maybe only necessary if you want to have a main.test.ts which isn't necessary when main is so small

.github/actions/create-ecr-repository/dist/157.index.js
hspitzley-czi, line 1: Why'd this directory change?

hspitzley-czi (approved):

.github/actions/terragrunt-provider-cache/package.json
jakeyheath: I'll defer to you. I feel like isolated packages.json's seems more reasonable and reducing the impact of changes across this repo

.github/actions/terragrunt-provider-cache/src/main.ts
jakeyheath: I'm also not a huge fan of having test code in the main code flow logic like that.

.github/actions/create-ecr-repository/dist/157.index.js
jakeyheath: I don't really know. That is actually what I need your help with. I don't really understand these ncc compiling artifacts the CI produces. Also, it seems like when it compiled is broken and the runs are failing. Here's a link:

https://github.com/chanzuckerberg/argus-infra-stacks/actions/runs/27581549542/job/81545052351

PR #550, opened 2026-06-12
feat: add reusable terragrunt-engine workflow (plan/apply)

jakeyheath:
## Overview

Part of generalizing the `argus-infra-stacks` Terragrunt CI into a reusable system other repos can adopt to replace Terraform Enterprise (proposal: #548). This adds the central engine: a `workflow_call` reusable workflow that a repo invokes with a small caller.

## Solution

`terragrunt-engine.yml` is the parity port of the argus `terragrunt.yml`. The plan/apply behavior is unchanged; the previously hardcoded values are now inputs:

- `ci_role_arn`, `cache_role_arn`, `provider_cache_bucket`, `aws_region`, `terraform_version`, `terragrunt_version`, `stack_paths`, `runner`.

Argus-specific behavior is opt-in so other repos are not forced into it:

- SSH agent and GitHub App token are auto-enabled from secret presence, so callers using `secrets: inherit` get them automatically.
- The SSH key and GitHub App credentials are declared optional secret inputs (`SHARED_INFRA_DEPLOY_KEY`, `EXTRA_DEPLOY_KEYS`, `GH_ACTIONS_HELPER_APP_ID`, `GH_ACTIONS_HELPER_PK`).
- Provider credentials follow naming conventions: `TF_ENV_<NAME>` secrets export as environment variables, `TF_VAR_<name>` secrets pass through as Terraform input variables.

The changed-stack detection is generalized to derive a stack from the configured base paths rather than hardcoded `terraform/envs` / `terraform/accounts`, while preserving the existing "skip brand-new stacks" behavior.

## Impact

No repo uses this engine yet, so nothing changes on merge.

This PR is **stacked on #549** (the composite actions it references): #549 is this PR's base branch, so review and merge #549 first and GitHub will retarget this PR to `main` automatically. The reference caller wiring is a separate cross-repo PR (chanzuckerberg/argus-infra-stacks#2573), which pins this engine by commit until it is released.

Part of #548 (PR2 of the delivery plan).


.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,325 @@
...
+            const fs = require("fs")
+            const read = (base) => fs.existsSync(base)
+              ? fs.readdirSync(base).map(f => `${base}/${f}`)
+              : []
+            const allChanges = [
+              ...new Set(read(process.env.ENVS_PATH)),
+              ...new Set(read(process.env.ACCOUNTS_PATH)),
+            ]
+            console.log(`allChanges: ${JSON.stringify(allChanges)}`)
+            core.setOutput("allChanges", allChanges)
```
jakeyheath: this terminology is too specific to fogg with envs and accounts path. it should be a configurable list of paths specified as on input to this workflow. The default should be the fogg ones (terraform/envs and terraform/accounts). Use the convention of having these paths split by newline string as the input.

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,325 @@
...
+        with:
+          initial-fetch-depth: "1"
+          list-files: json
+          filters: |
+            changed:
+              - added|modified: '${{ inputs.tf_envs_path }}/**'
+              - added|modified: '${{ inputs.tf_accounts_path }}/**'
+            added_makefiles:
+              - added: '${{ inputs.tf_envs_path }}/**/Makefile'
+              - added: '${{ inputs.tf_accounts_path }}/**/Makefile'
```
jakeyheath: the added makefiles was a hack to not try to terraform plan things that were new. This isn't really needed anymore now that we have changed the convention to apply on PR. It's also very fogg specific.

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in b689e04. Replaced `tf_envs_path`/`tf_accounts_path` with a single newline-separated `stack_paths` input, defaulting to the fogg layout (`terraform/envs` and `terraform/accounts`). The all-stacks enumeration, the paths-filter filters, and the changed-stack detection are all generated from that list, and artifact naming now uses `basename` instead of assuming a 3-segment path.

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,325 @@
...
+        run: |
+          mkdir -p ~/.ssh/
+          ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+      - name: Install SSH key
+        if: ${{ inputs.use_ssh_agent }}
+        uses: webfactory/ssh-agent@v0.10.0
+        with:
+          ssh-private-key: ${{ secrets.ssh_private_key }}
+      - name: Generate token
+        id: generate_token
```
jakeyheath, line 145: add a TODO here that we need to fix this. It will be annoying to setup ssh keys for each repo just so they can download terraform modules

jakeyheath: can you provide some suggestions in the code as comments? I will revisit them later

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in 1750cb4. Removed the `added_makefiles` filter and the new-stack skip entirely, so the engine no longer special-cases newly added stacks. As you note, that hack was fogg-specific and unnecessary under apply-on-PR.

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,325 @@
...
+      tf_accounts_path:
+        description: "Base path for per-account stacks."
+        required: false
+        type: string
+        default: terraform/accounts
+      create_module_symlinks:
+        description: "Create per-component `modules` symlinks (fogg + Terragrunt 1.0 workaround). Off for non-fogg repos."
+        required: false
+        type: boolean
+        default: false
```
jakeyheath: I feel like this should just always be turned on and probably not an input even for things that aren't fogg repos

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,325 @@
...
+          cache: enable
+      - name: Install Terraform
+        uses: hashicorp/setup-terraform@v4
+        with:
+          terraform_wrapper: false
+          terraform_version: ${{ inputs.terraform_version }}
+      - name: Install AWS CLI
+        uses: chanzuckerberg/github-actions/.github/actions/install-aws-cli@main
+      - name: Create local module symlinks
+        if: ${{ inputs.create_module_symlinks }}
```
jakeyheath: I think this should always be run. See comment above

.github/workflows/terragrunt-engine.yml
jakeyheath: Added a TODO above the Install SSH key step in 7754a80 calling out that per-repo SSH deploy keys for private module access are painful, and that we should replace it (e.g. a GitHub App token scoped to the module repos, or a shared modules source).

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,325 @@
...
+          aws-region: ${{ inputs.aws_region }}
+          role-to-assume: ${{ inputs.cache_role_arn }}
+          role-chaining: true
+          role-session-name: TerragruntEngineCache
+      - name: Restore Provider Cache from S3
+        uses: chanzuckerberg/github-actions/.github/actions/terragrunt-provider-cache@main
+        with:
+          operation: restore
+          provider-cache-bucket: ${{ inputs.provider_cache_bucket }}
+          provider-cache-key: ${{ inputs.provider_cache_key }}
```
jakeyheath: The provider bucket will always be the same, set it as a default input. The key should be inferred from the calling repo. Right now, it looks like this:

terraform/argus-infra-stacks/provider-cache/providers.tar.gz

But everyone should essentially be the same, except for the owner and repo. It would be good to update it so it was

terraform/<gh_owner>/<gh_repo>/provider-cache/providers.tar.gz

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in c74f3cd. Removed the `create_module_symlinks` input and made the step always run, with a guard so it no-ops on repos without a `terraform/modules` directory. (The argus caller will drop the now-removed input when it bumps from its pinned SHA to the released engine.)

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in c74f3cd. The module-symlink step is no longer gated on an input; it always runs and no-ops when there is no `terraform/modules` dir. See the reply on the input definition above.

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,325 @@
...
+          aws-region: ${{ inputs.aws_region }}
+          role-to-assume: ${{ inputs.ci_role_arn }}
+          role-session-name: TerragruntEngine
+          unset-current-credentials: true
+      - name: Terragrunt ${{ inputs.action }}
+        working-directory: ${{ matrix.tfmodule }}
+        env:
+          OKTA_BASE_URL: ${{ secrets.okta_base_url }}
+          OKTA_ORG_NAME: ${{ secrets.okta_org_name }}
+          OKTA_API_TOKEN: ${{ secrets.okta_api_token }}
```
jakeyheath: this is an argus-infra-stacks thing. I think we didn't resolve this in the plan. Each repo and workspace might need sets of credentials to be added to the environment for the various providers running. Where should we store these? I will return to the plan to address that concern, but for now, we shouldn't have the Okta env variables as part of the generic action. If you have suggestions, update the plan and include them in the open questions section of the proposal when you are addressing this comment.

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,333 @@
...
+        default: us-west-2
+      ci_role_arn:
+        description: "IAM role the engine assumes to run Terragrunt (OIDC). Defaults to the fleet-wide argus-infra-stacks CI role."
+        required: false
+        type: string
+        default: arn:aws:iam::416703108729:role/argus-infra-stacks-ci
+      cache_role_arn:
+        description: "IAM role (assumed via role chaining) for the provider-cache S3 bucket and lock cleanup."
+        required: true
+        type: string
```
jakeyheath, line 40: this should have a default. use whatever is in argus-infra-stacks. These roles will be used by everyone most likely

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,333 @@
...
+      provider_cache_bucket:
+        description: "S3 bucket holding the shared Terraform provider cache tarball. Defaults to the fleet-wide argus bucket."
+        required: false
+        type: string
+        default: core-platform-infra-prod-prod-argus-infra-stacks-state
+      provider_cache_key:
+        description: "S3 object key for the provider cache tarball. Defaults to the fleet-wide argus key."
+        required: false
+        type: string
+        default: terraform/argus-infra-stacks/provider-cache/providers.tar.gz
```
jakeyheath: remove this as an input. We should be deriving the key by gh_owner and gh_repo using this format string (see comment below)

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in 92c86a1. The bucket stays a default input, and the key is now inferred from the calling repo: it defaults to `terraform/<owner>/<repo>/provider-cache/providers.tar.gz` via `github.repository` (a caller can still override `provider_cache_key`). Note this changes argus's key from `terraform/argus-infra-stacks/...` to `terraform/chanzuckerberg/argus-infra-stacks/...`, so the first run repopulates the cache at the new key.

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in 92c86a1. Removed the Okta secrets and env from the generic engine. I added an **Open questions** section to the proposal (#548) covering where per-repo/per-workspace provider credentials should live and how the engine should inject them, with a leaning toward fetching a provider-env blob from AWS Secrets Manager/SSM at runtime (keyed by owner/repo/stack) since the engine already holds the CI role. Argus keeps working by pinning an engine commit that still includes Okta until we settle this.

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in ba5ff3b. `cache_role_arn` now defaults to the fleet-wide argus tfe-si role (`arn:aws:iam::533267185808:role/tfe-si`) and is no longer required.

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in ba5ff3b. Removed `provider_cache_key` as an input entirely; the key is always derived as `terraform/${{ github.repository }}/provider-cache/providers.tar.gz` (i.e. terraform/<owner>/<repo>/provider-cache/providers.tar.gz).

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,251 @@
...
+    strategy:
+      fail-fast: false
+      matrix:
+        tfmodule: ${{ fromJson(needs.find-changed-dirs.outputs.allChanges) }}
+    if: ${{ needs.find-changed-dirs.outputs.allChanges != '[]' }}
+    env:
+      TG_PROVIDER_CACHE: "1"
+      TG_PROVIDER_CACHE_DIR: /tmp/terragrunt-provider-cache
+      # Cache key is always derived from the calling repo: terraform/<owner>/<repo>/...
+      PROVIDER_CACHE_KEY: ${{ format('terraform/{0}/provider-cache/providers.tar.gz', github.repository) }}
```
jakeyheath, line 125: include the owner and the repo as part of the key

jakeyheath: `github.repository` already expands to `<owner>/<repo>`, so the derived key is `terraform/<owner>/<repo>/provider-cache/providers.tar.gz` (e.g. `terraform/chanzuckerberg/argus-infra-stacks/provider-cache/providers.tar.gz`) — owner and repo are both included. Happy to split it into explicit `github.repository_owner` + repo name if you'd prefer it spelled out.

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,251 @@
...
+        uses: actions/checkout@v6
+        with:
+          ref: ${{ steps.ref.outputs.headRef || 'refs/heads/main' }}
+          token: ${{ steps.generate_token.outputs.token || github.token }}
+      - name: Configure AWS credentials (CI base role)
+        uses: aws-actions/configure-aws-credentials@v6
+        with:
+          aws-region: ${{ inputs.aws_region }}
+          role-to-assume: ${{ inputs.ci_role_arn }}
+          role-session-name: TerragruntEngine
```
jakeyheath, line 168: is this used by anything? A few steps down we assume the cache role, then we forget that role and assume this role again. Maybe this whole step can be removed

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,251 @@
...
+        description: "Load ssh_private_key into an ssh-agent (for private module sources)."
+        required: false
+        type: boolean
+        default: false
+      use_github_app_token:
+        description: "Generate a GitHub App token (gh_app_id + gh_app_private_key) for checkout."
+        required: false
+        type: boolean
+        default: false
+    secrets:
```
jakeyheath, line 73: for provider credentials, can we add a inputs section that a user can provider a set of secrets as a newline separated list and then those values are added as environment variables to the terragrunt run?

.github/workflows/terragrunt-engine.yml
jakeyheath: github.repository is fine. I didn't know it included owner

.github/workflows/terragrunt-engine.yml
jakeyheath: It is used: the next AWS step assumes the cache role with `role-chaining: true`, which assumes that role FROM the credentials this step establishes via OIDC. The cache role (tfe-si) is only assumable by chaining from the CI role, not via GitHub OIDC directly, so the CI role has to be in the environment first. The later re-assume (with `unset-current-credentials: true`) is the intentional return to the CI role for the Terragrunt run after the provider-cache detour. So the sequence is OIDC -> CI role -> chain to cache role (restore cache) -> back to CI role (terragrunt). Removing this step would break the role-chaining. I added a comment in 7ee5e9d making that dependency explicit.

.github/workflows/terragrunt-engine.yml
jakeyheath: Done in 793a6e2. Added a `provider_env` secret: a newline-separated list of `NAME=value` pairs that the engine exports as environment variables (each value masked via `::add-mask::`) right before the Terragrunt run. It is provider-agnostic — a caller supplies whatever it needs (argus would pass its OKTA_* values) without the engine knowing about any specific provider. Updated the Open questions section of the proposal (#548) to record this as the chosen approach, keeping a runtime SSM/Secrets Manager fetch noted as a future option for per-workspace scoping.

.github/actions/create-ecr-repository/dist/157.index.js
jakeyheath, line 5: why are there all these js and js map files?

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,282 @@
...
+  find-changed-dirs:
+    runs-on: ${{ inputs.runner }}
+    outputs:
+      allChanges: ${{ steps.stacks.outputs.stacks }}
+    timeout-minutes: 10
+    steps:
+      - uses: actions/checkout@v6
+      - name: Resolve stacks
+        id: stacks
+        uses: chanzuckerberg/github-actions/.github/actions/find-changed-stacks@main
```
jakeyheath: pin to latest major tag

jakeyheath: maybe after we have merged it. add a todo for now

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,282 @@
...
+          stack_paths: ${{ inputs.stack_paths }}
+          all_stacks: ${{ inputs.all_stacks }}
+
+  run-all:
+    runs-on: ${{ inputs.runner }}
+    needs: find-changed-dirs
+    strategy:
+      fail-fast: false
+      matrix:
+        tfmodule: ${{ fromJson(needs.find-changed-dirs.outputs.allChanges) }}
```
jakeyheath: what is the limit on matrix? What happens if the number of stacks exceeds it?

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,282 @@
...
+      # or a shared modules source) so onboarding does not need per-repo SSH keys.
+      - name: Install SSH key
+        if: ${{ steps.creds.outputs.ssh == 'true' }}
+        uses: webfactory/ssh-agent@v0.10.0
+        with:
+          ssh-private-key: ${{ secrets.ssh_private_key }}
+      - name: Generate token
+        id: generate_token
+        if: ${{ steps.creds.outputs.app == 'true' }}
+        uses: actions/create-github-app-token@v3
```
jakeyheath, line 146: do any of these actions need updating to the latest major tag?

.github/workflows/terragrunt-engine.yml
jakeyheath: Added a TODO in 0b48afb above these references noting to pin all the chanzuckerberg/github-actions action refs (find-changed-stacks, install-aws-cli, terragrunt-provider-cache, release-terraform-locks) to a released major tag once this merges, instead of @main.

.github/actions/create-ecr-repository/dist/157.index.js
jakeyheath: These are committed ncc build artifacts (this repo commits each JS action's dist/). Two separate things:

- **find-changed-stacks/dist/*** — the bundle for the new action this PR adds (expected).
- **create-ecr-repository/dist/*** (the 157/186/... files are aws-sdk lazy-load chunks) — this PR's new workspaces add @aws-sdk/client-dynamodb (for release-terraform-locks), which bumped a shared @aws-sdk/@smithy transitive dependency in the lockfile, so the repo's CI (the 'ci - fix linting / update build artifacts' commit) rebuilt create-ecr-repository's bundle to match. I confirmed it is reproducible from the current lockfile — reverting it just gets regenerated on the next CI run — so it can't be cleanly dropped while these deps are present.

How would you like to handle it? (a) leave it as CI keeping artifacts in sync; (b) I dig into the exact transitive bump and pin it so create-ecr is not perturbed; (c) accept it as a one-time artifact refresh. Leaving this open for your call.

.github/workflows/terragrunt-engine.yml
jakeyheath: GitHub caps a matrix at 256 jobs per workflow run. If the changed-stack count exceeded 256, the matrix would fail to expand. In practice that is not a concern: the per-PR changed set is small, and even the nightly all-stacks run (argus has roughly 50 envs/accounts) is well under 256. If a repo ever approached the limit, we would batch stacks into chunks (matrix over groups, terragrunt run --all within each). Not an issue at current scale.

.github/workflows/terragrunt-engine.yml
jakeyheath: Checked the latest releases — every third-party action pin is already on its latest major:
- actions/checkout@v6 (latest v6.0.3)
- actions/create-github-app-token@v3 (v3.2.0)
- actions/upload-artifact@v7 (v7.0.1)
- hashicorp/setup-terraform@v4 (v4.0.1)
- aws-actions/configure-aws-credentials@v6 (v6.2.0)
- webfactory/ssh-agent@v0.10.0 (latest)
- jaxxstorm/action-install-gh-release@v3.0.0 (latest)

None need a major bump.

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,285 @@
...
+        required: false
+        type: string
+        default: |
+          terraform/envs
+          terraform/accounts
+    # Secrets are passed with `secrets: inherit`. The engine reads these org-wide
+    # secrets by naming convention, so callers need no per-secret mapping. Each is
+    # optional; the related step is skipped when its secret is absent:
+    #   GH_ACTIONS_HELPER_APP_ID / GH_ACTIONS_HELPER_PK - GitHub App token for checkout
+    #   TERRAGRUNT_SSH_PRIVATE_KEY                       - ssh-agent key for private modules
```
jakeyheath: this should be called "SHARED_INFRA_DEPLOY_KEY"

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,285 @@
...
+          ssh-keyscan -H github.com >> ~/.ssh/known_hosts
+      # TODO: requiring a per-repo SSH deploy key just to fetch private Terraform
+      # modules is painful to set up for every adopting repo. Replace this with a
+      # better mechanism (e.g. a GitHub App token scoped to read the module repos,
+      # or a shared modules source) so onboarding does not need per-repo SSH keys.
+      - name: Install SSH key
+        if: ${{ steps.creds.outputs.ssh == 'true' }}
+        uses: webfactory/ssh-agent@v0.10.0
+        with:
+          ssh-private-key: ${{ secrets.TERRAGRUNT_SSH_PRIVATE_KEY }}
```
jakeyheath: add an input for users to be able to add additional ssh keys per repo. For example, in argus-infra-stacks, we add ${{ secrets.CORE_PLATFORM_INFRA_DEPLOY_KEY }}, but that key isn't needed by all repos. 

I believe the ssh-private-key argument takes newline separted private keys here

.github/workflows/terragrunt-engine.yml
jakeyheath: Renamed the secret to `SHARED_INFRA_DEPLOY_KEY` (doc comment + both usages in the run-all job).

.github/workflows/terragrunt-engine.yml
jakeyheath: Added an optional `EXTRA_DEPLOY_KEYS` convention secret (newline-separated PEMs) that's appended to the shared key and loaded by ssh-agent, so a repo like argus-infra-stacks can supply `CORE_PLATFORM_INFRA_DEPLOY_KEY` without every repo needing it. The detect-credentials step now enables ssh when either secret is present.

.github/actions/find-changed-terragrunt-stacks/action.yml
hspitzley-czi, line 1: recommend renaming to `find-changed-tf-stacks` or `find-changed-terragrunt-stacks` to disambiguate with other places we use the term `stack`

.github/actions/find-changed-stacks/src/main.ts
```diff
@@ -0,0 +1,67 @@
+import * as core from '@actions/core';
+import * as github from '@actions/github';
+import * as fs from 'fs';
+import {
+  enumerateStacks,
+  parseBases,
+  stacksFromChangedFiles,
+} from './lib';
+
+async function changedFiles(token: string): Promise<string[]> {
```
hspitzley-czi: can we reuse https://github.com/chanzuckerberg/github-actions/blob/d2e0764f8e895d00e1b59d83046f64c00392e0f3/.github/actions/find-changed-files/src/findChangedFiles.ts#L9 instead of reimplementing?

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,309 @@
...
+        required: false
+      EXTRA_DEPLOY_KEYS:
+        description: "Additional ssh-agent keys (newline-separated PEMs) for repos needing more than the shared key."
+        required: false
+
+permissions:
+  actions: write
+  contents: write
+  id-token: write
+  # Required for dorny/paths-filter listing PR files on pull_request (read-only).
```
hspitzley-czi: stale. dorny isn't used in this workflow anymore

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,309 @@
...
+          stack_paths: ${{ inputs.stack_paths }}
+          all_stacks: ${{ inputs.all_stacks }}
+
+  run-all:
+    runs-on: ${{ inputs.runner }}
+    needs: find-changed-dirs
+    strategy:
+      fail-fast: false
+      matrix:
+        tfmodule: ${{ fromJson(needs.find-changed-dirs.outputs.allChanges) }}
```
hspitzley-czi: nit: this is a matrix over stacks, not modules

.github/actions/find-changed-stacks/src/main.ts
```diff
@@ -0,0 +1,67 @@
...
+  return [];
+}
+
+export async function run(): Promise<void> {
+  const bases = parseBases(core.getInput('stack_paths', { required: true }));
+  const allStacks = core.getBooleanInput('all_stacks');
+
+  let stacks: string[];
+  if (allStacks) {
+    stacks = enumerateStacks(bases, (base) => (fs.existsSync(base) ? fs.readdirSync(base) : null));
```
hspitzley-czi: probably shouldn't be an issue based on how we do things but `readdirSync` will include files, not just dirs, so if there were a file under `base` it would be treated as a stack

.github/workflows/terragrunt-engine.yml
```diff
@@ -0,0 +1,309 @@
...
+      - name: Generate token
+        id: generate_token
+        if: ${{ steps.creds.outputs.app == 'true' }}
+        uses: actions/create-github-app-token@v3
+        with:
+          client-id: ${{ secrets.GH_ACTIONS_HELPER_APP_ID }}
+          private-key: ${{ secrets.GH_ACTIONS_HELPER_PK }}
+      - name: Get Base Git Ref
+        id: ref
+        if: ${{ inputs.all_stacks != true }}
```
hspitzley-czi: I'm a bit confused by this guard, it looks like it's there to force `steps.ref.outputs.headRef` to be null so we fallback to `refs/heads/main`, but I don't get why we want to fallback to main. If this workflow is invoked from main won't `get-github-ref-names` output `main` as the head ref?

jakeyheath: That all_stacks is mainly reserved for cronjobs and apply the terraform across the whole repo. It isn't designed to work on a PR, just the main branch. I'm not sure why I had that guard in the original argus-infra-stacks, though because I think you are right that its unnecessary.

PR #568, opened 2026-06-26
feat: add stack-authority action and generalize backend discovery

jakeyheath:
## Overview

The apply-on-PR system needs a durable per-stack authority record that tracks which ref (main or a PR) currently owns a stack. This prevents cross-PR clobbering and stops the nightly drift job from reverting unmerged work. Native Terraform state locks are ephemeral and carry no durable owner/ref, so they can't fill this role.

## Solution

Adds the `stack-authority` composite action with four operations:

- **read** — look up who owns a set of stacks (main or a specific PR)
- **claim** — atomically take ownership for a PR using a DynamoDB conditional write that only succeeds if the stack is currently main-owned (or already owned by the same PR)
- **release** — reset ownership to main (deletes the authority row)
- **reap** — scan for stale claims older than a configurable age and release them

Authority rows live in the same DynamoDB table as Terraform state locks, namespaced under `authority/<repo>/<stack>` so they never collide with native lock rows. No new infrastructure is needed.

Also generalizes backend discovery in `release-terraform-locks` so it works on non-fogg repos. Previously it only looked for `fogg.tf` files. Now it checks `fogg.tf`, `backend.tf`, then falls back to any `.tf` file containing a `backend "s3"` block. The old `findFoggTfFiles` function is preserved as a deprecated alias.

## Impact

- The authority action is inert until wired into the engine (PR5). Existing callers are unaffected.
- Lock cleanup now works on repos that use `backend.tf` or other file names for their S3 backend configuration.

Part of the apply-on-PR effort ([#548](https://github.com/chanzuckerberg/github-actions/issues/548)). Stacked: this lands before PR5 (wire authority into engine).


.github/actions/stack-authority/.eslintrc.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  rules: {
+    'no-restricted-syntax': 'off',
+    'no-continue': 'off',
```
hspitzley-czi: do we need this new file or can we put these in the root .eslintrc.js so the linting is consistent across all actions?

.github/actions/stack-authority/src/index.ts
```diff
@@ -0,0 +1,148 @@
...
+      if (!allClaimed) {
+        core.setFailed(
+          'One or more stacks could not be claimed (owned by another PR)',
+        );
+      }
+      break;
+    }
+
+    case 'release': {
+      if (stacks.length === 0 && !prNumber) {
```
hspitzley-czi, line 133: this validates that either stacks or pr number is provided but then doesn't use pr number, is that expected?

.github/actions/stack-authority/src/index.ts
```diff
@@ -0,0 +1,148 @@
...
+          results[stack] = result.record;
+        } else {
+          results[stack] = { ownerRef: 'main' };
+        }
+        const rec = result.record;
+        const owner = rec?.ownerRef ?? 'main';
+        const ownerLabel = owner === 'main' ? 'main' : `PR #${rec?.prNumber}`;
+        core.info(`${stack}: owned by ${ownerLabel}`);
+      }
+      core.setOutput('results', JSON.stringify(results));
```
hspitzley-czi, line 82: this outputs camel case but the .github/actions/stack-authority/action.yml output says it is snakecase

.github/actions/stack-authority/src/lib.ts
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * Uses a DynamoDB Scan with a FilterExpression on the authority/<repo>/ prefix.
+ * The state-lock table's partition key is LockID (string) with no sort key,
+ * so Query with begins_with is not possible. Since authority rows are sparse
+ * (one per stack per repo, typically tens of rows in the whole table), a
+ * filtered scan is acceptable.
+ *
+ * Note: the caller is responsible for checking whether the owning PR is
+ * closed (via the GitHub API). This function only checks age. The index.ts
+ * orchestrator handles the PR-closed check.
```
hspitzley-czi: nit: the `reap` case in index.ts doesn't check if the PR is closed before calling `reap()`. is that expected?

.github/actions/stack-authority/src/index.ts
hspitzley-czi: nvm, i see it's changed to be consistent in https://github.com/chanzuckerberg/github-actions/pull/569/changes

hspitzley-czi (approved):

PR #583, opened 2026-06-30
feat(terragrunt-engine): enforce branch protection during apply validation

jakeyheath:
## Overview

The terragrunt engine requires specific repo settings and branch protections to function correctly, but there's no mechanism to ensure these are in place. Repos can silently misconfigure settings (disable auto-merge, allow non-squash merges, omit the required status check), breaking the apply-and-merge flow without any feedback.

## Solution

A new `terragrunt-branch-protection` action runs during `/apply-and-merge` validation and auto-fixes any drift in three repo settings:

- **Auto-merge** is enabled (required for the apply-then-auto-merge flow).
- **Squash merge** is the only allowed merge method (merge commits and rebase merges are disabled).
- **The apply-gate status check** (default: `terragrunt-apply`) is required on the default branch.

The action uses a single GraphQL query to read all state, then patches only what needs fixing via REST (repo settings) and GraphQL mutations (branch protection rules). It preserves existing branch protection settings when adding the required check.

The engine's `validate-comment` job now generates a GitHub App token after a successful `/apply-and-merge` validation and runs the branch protection check. This also fixes the finalize auto-merge to use `SQUASH` instead of `MERGE`, since squash-only is now enforced.

## Impact

Repos using the terragrunt engine will have their branch protection auto-corrected on every apply, so misconfiguration is self-healing rather than a silent failure mode.

## Related

- chanzuckerberg/argus-infra-stacks#2675 (the apply-on-PR cutover that depends on these protections)
- chanzuckerberg/github-actions#575 (engine finalize)


jakeyheath:
Subsumed by #585 which includes this change plus all the other engine fixes.

PR #598, opened 2026-07-21
feat(argus-builder): add per-image zstd layer compression

jakeyheath:
## Overview

Large container images (10+ GB) spend most of their CI time on layer compression and upload. Switching from gzip to zstd cuts build time by up to 74% and image size by ~17% (measured in [evolutionaryscale#12712](https://github.com/evolutionaryscale/evolutionaryscale/pull/12712)).

This adds opt-in, per-image zstd layer compression to the shared argus-docker-build pipeline so any consumer can enable it without forking the action.

## Solution

Three new optional fields on the per-image JSON object (`compression`, `compression_level`, `force_compression`) are threaded through the full build chain:

- `build-prep` validates them against the schema and passes them into the matrix.
- `argus-docker-build.yaml` forwards them from the matrix to `argus-builder/docker-build`.
- `argus-builder/docker-build` forwards them to `docker-build-push`.
- `docker-build-push` computes an explicit `--output` exporter string (`type=registry,compression=zstd,...`), replacing the previous `push: true`.
- `argus-builder-dispatch.yaml` forwards them from the resolved matrix so the orchestrator path works too.

Defaults are `gzip` / level `3` / `force_compression: false`, matching docker's existing behavior, so no consumer changes.

## Impact

Any Argus-deployed service can opt into zstd by adding three keys to its image definition:

```json
{
  "compression": "zstd",
  "compression_level": "3",
  "force_compression": true
}
```

This works for both callers of the reusable workflow and callers of the composite action directly.

## References

- [CCIE-7002](https://czi.atlassian.net/browse/CCIE-7002)
- [evolutionaryscale#12712](https://github.com/evolutionaryscale/evolutionaryscale/pull/12712) -- zstd compression for Forge images (74% build time reduction, 17% size reduction)
- [Docker image exporter docs](https://docs.docker.com/build/exporters/image-registry/) -- compression, compression-level, force-compression parameters
- [argus-example-app#442](https://github.com/chanzuckerberg/argus-example-app/pull/442) -- test PR validating zstd on frontend image vs gzip on the other three
- [argus#2597](https://github.com/chanzuckerberg/argus/pull/2597) -- CLI/config side: adds compression fields to resolve-matrix output

[CCIE-7002]: https://czi.atlassian.net/browse/CCIE-7002?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


imathur1 (approved):
thank you!

.github/workflows/argus-docker-build.yaml
```diff
@@ -130,6 +130,9 @@ jobs:
           image_tag: |
             ${{ needs.prep.outputs.image_tag }}
             ${{ contains(inputs.envs, 'prod') && format('prod-{0}', needs.prep.outputs.image_tag) || '' }}
+          compression: ${{ matrix.image.compression }}
+          compression_level: ${{ matrix.image.compression_level }}
+          force_compression: ${{ matrix.image.force_compression }}
```
hspitzley-czi, line 135: This won't do much unless woven into the `resolve-matrix` command in the orchestrator too

jakeyheath: isn't the reusable workflow still used in the orchestrator?

hspitzley-czi: nope, orchestrator uses https://github.com/chanzuckerberg/github-actions/blob/main/.github/workflows/argus-builder-dispatch.yaml

jakeyheath: oo right. ok, I will add it there as well

hspitzley-czi:
> Added the forwarding in [046bce3](https://github.com/chanzuckerberg/github-actions/commit/046bce30214819b8d0df218c01f6d61f9bcbba5e). The dispatch workflow now passes compression, compression_level, and force_compression from the matrix to docker-build. Still need a CLI change to make resolve-matrix emit those fields from .argus-ci.yaml.

@jakeyheath once you have made the CLI changes, use https://github.com/chanzuckerberg/argus/tree/main/.cursor/skills/test-ci-orchestrator to test the changes

.github/workflows/argus-docker-build.yaml
jakeyheath: @hspitzley-czi why do we install go and build argus from source? I thought that was an artifact of testing on a branch.

Also, do you have any problems with making this new compression style the default. It seems to be accurate that it produces smaller compression layers

hspitzley-czi: Install/build from source is only if you specify argus_ref https://github.com/chanzuckerberg/github-actions/blob/23e8645ce5c82d854da3a1148eb3a9d185f46084/.github/workflows/argus-builder-dispatch.yaml#L65

This is to allow testing changes to the CLI without making a branch/PR in github-actions repo

hspitzley-czi: I'm fine with making it the default if it's thoroughly tested

jakeyheath:
E2e test via orchestrator path passed. Used the [test-ci-orchestrator skill](https://github.com/chanzuckerberg/argus/tree/main/.cursor/skills/test-ci-orchestrator).

- Deployed argus@heathj/zstd-compression as rdev stack `argus-rdev-f3842a2e`
- Triggered `sync-ci-webhooks` to point argus-webhook-app's webhook at the rdev stack
- Pushed to argus-webhook-app with `compression: zstd` on the frontend image and `argus_ref: heathj/zstd-compression`
- The dispatched build (run [29938897828](https://github.com/chanzuckerberg/argus-webhook-app/actions/runs/29938897828)) succeeded:
  - `resolve-matrix` emitted `"compression": "zstd"` in the matrix JSON
  - frontend built with `--output type=registry,compression=zstd,compression-level=3,force-compression=true`
  - other images built with `--output type=registry,compression=gzip,compression-level=3,force-compression=false`

CLI-side PR: https://github.com/chanzuckerberg/argus/pull/2597

.github/actions/docker-build-push/action.yml
```diff
@@ -102,12 +114,24 @@ runs:
...
+        c="${COMPRESSION:-gzip}"
+        cl="${COMPRESSION_LEVEL:-3}"
+        fc="${FORCE_COMPRESSION:-true}"
+        echo "spec=type=registry,compression=${c},compression-level=${cl},force-compression=${fc}" >> "$GITHUB_OUTPUT"
     - name: Build and push
       uses: docker/build-push-action@v7
       with:
         file: ${{ inputs.dockerfile }}
         context: ${{ inputs.context }}
-        push: true
```
hspitzley-czi, line 110: is this removal intentional?

jakeyheath: yes, I confirmed that. I guess the action doesn't support an input like compression, so you have to set it on the output. The outputs formatting now includes with push true did before

hspitzley-czi: ohhh, i see. nice

hspitzley-czi (commented):
Approved once branch refs are set back to `v6`

hspitzley-czi (approved):

PR #602, opened 2026-08-03
feat(codeowners-approval-check): add CODEOWNERS approval status check action

hspitzley-czi:
Add a node24 action that passes only when every changed file matching a CODEOWNERS rule is authored or approved by a code owner. Team owners are expanded to members via an org-scoped token. CODEOWNERS is read from the PR base ref to prevent self-approval bypass.


alexlokshin-czi (approved):

jakeyheath:
Isn't this what a branch protection for codeowners does?

hspitzley-czi:
> Isn't this what a branch protection for codeowners does?

@jakeyheath The "Require approval from codeowners" branch protection setting is causing issues in the evoscale monorepo. When there are only 1-2 codeowners and the PR author is one of them we want to require an approval but not necessarily from the other codeowner (they may be OOO or just busy). Github branch protections doesn't have this flexibility of allowing code-owning authors to bypass the required codeowner approval

jakeyheath:
> > Isn't this what a branch protection for codeowners does?
> 
> @jakeyheath The "Require approval from codeowners" branch protection setting is causing issues in the evoscale monorepo. When there are only 1-2 codeowners and the PR author is one of them we want to require an approval but not necessarily from the other codeowner (they may be OOO or just busy). Github branch protections doesn't have this flexibility of allowing code-owning authors to bypass the required codeowner approval

Ah gotcha. How many github actions do we have that circumvent annoying github stuff? 🤦🏻

jakeyheath (approved):

